PR #791, opened 2014-04-14
WIP - Add Debian init script

bracki:
Hi,

this adds a Debian init.d script. In its current state in it's only useful for _real_ Debian distros.
Also it has some customizations like bouncing the version and a hardcoded `$GOPATH`, because I couldn't get it to build on `Squeeze` otherwise.
In order to support Ubuntu, we would have to add an upstart script conditionally instead of the systemv one.
I also moved up the `DEB` block in https://github.com/Jimdo/heka/blob/3ad99212f28b9bd5a251003f056fc26e8dc94a71/CMakeLists.txt to be before the `include(CPack)`. Otherwise the `CPACK_DEBIAN_*` vars are not picked up properly.

### DIFF
--- a/debian/postrm
+++ b/debian/postrm
@@ -1,9 +1,14 @@
 #!/bin/sh
 
+USER="heka"
+GROUP="$USER"
+
 set -e
 
 if [ "$1" = "purge" ]
 then
 	update-rc.d heka remove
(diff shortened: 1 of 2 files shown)


nickchappell:
+1

PR #1078, opened 2014-09-17
WIP Handle SIGHUP

ecnahc515:
Here's a work in progress of handling SIGHUP gracefully.

Currently test's aren't updated with the new code, so they will fail. I'm looking for review because there are some rough areas (in particular all the mutexes).

It will keep open TCP Input's listeners and connections upon a SIGHUP, and then return them after heka re-initializes. Any listeners which no longer have a plugin which matches the current config will be closed after re-initializing plugins.


robbiev:
Is this no longer happening?

rafrombrc:
It's not completely abandoned, but it's not the highest priority for us right now, and it sems that @ecnahc515 no longer has time to work on it, so it's being tabled for now.

robbiev:
Sure I understand. Allow me to just dump my use case here as a +1

Currently I'm using heka with docker and the Logstreamer input. Each time I start a container I generate a heka config file for that container and then restart heka. This is where the need for SIGHUP comes from - it would be nice to not have to restart heka - currently I send SIGTERM and then let supervisord revive it. I assume that SIGHUP would allow for a quicker reload (with very minimal downtime in theory) and thus result in a better user experience at the receiving end when viewing logs as they come in (a shorter or hardly noticeable interruption time because of heka restart).

I know DockerLogInput exists but IIRC it is possible to miss logs when heka is not active (e.g. when restarting heka, after a crash or while booting the machine) and also at the very start when a container boots I think there is a potential time window in which container logs could go missing. I don't want to miss any logs if I can help it so I came up with the workaround I explained.

DanyC97:
I see this as being closed, has this been implemented on any version recent version?

PR #1248, opened 2014-12-27
Handle multi-line statsd input more like statsd

wyndhblb:
Some statsd emitters put a double \n\n in the output of multi line statsd stats.  This causes the entire message to be treated as failed, when there is an easy fix here.  Same behavior as statsd itself https://github.com/etsy/statsd/blob/master/stats.js line  205 or so.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -80,7 +80,6 @@ Features
   (more like statsd itself).
   
 
->>>>>>> upstream/dev
 0.8.3 (2015-01-08)
 ==================
 
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -61,6 +61,11 @@ Features
 
 * Added `send_interval` setting to SmtpOutput.
 
+* Added ability for Statsd input handler to treat a malformed stat line 
+  one at time without skipping all the "good" stats in a multi-line input
+  (more like statsd itself).
+  
+
 0.8.1 (2014-12-17)
 ==================
 
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -13,13 +13,11 @@ Backwards Incompatibilities
   `send_decode_failures` config options, automatically extracted by Heka's
   config system.
 
-* InputRunner interface has added a `Deliver` method, which handles decoding
-  and injection to the router according to the specifications of the
-  `decoder`, `synchronous_decode`, and send_decode_failure` config options,
-  *except* in cases when an input implements a `SetCommonInputConfig` method.
-  If that is implemented, the InputRunner will pass in a struct containing the
-  specified decoding options, and the input will be expected to handle
-  decoding accordingly.
+* InputRunner now handles message decoding and delivery to the router,
+  according to the specifications of the input's configuration. This is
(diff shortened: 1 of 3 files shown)


plugins/statsd/statsd_input.go
```diff
@@ -125,20 +125,24 @@ func (s *StatsdInput) Stop() {
...
 	}
 
+	if warn != nil || len(warn) > 0 {
+		s.ir.LogError(fmt.Errorf("Warning Part of message could not be parsed: %s", warn))
+	}
+
 	for _, stat := range stats {
 		if !s.statAccum.DropStat(stat) {
-			s.ir.LogError(fmt.Errorf("Undelivered stat: %v", stat))
+			s.ir.LogError(fmt.Errorf("Undelivered stat: %s", stat))
```
rafrombrc: %v will output the stat values, converting to string where necessary, whereas %s will generate an error for non-string attributes of the struct. If we're going to change it at all, using %+v to include the field names is a better choice.

plugins/statsd/statsd_input.go
```diff
@@ -152,15 +156,34 @@ func parseMessage(message []byte) ([]Stat, error) {
 		lines = [][]byte{message}
 	}
 
-	for _, line := range lines {
+	var num_lines int = len(lines)
+	var bad_lines_ct int = 0
```
rafrombrc: We don't really need to keep a separate count of the bad lines, do we? We can just check len(bad_lines) instead.

plugins/statsd/statsd_input_test.go
```diff
@@ -166,7 +166,7 @@ func TestParseMessage(t *testing.T) {
 	}
 
 	for msg, expected := range testData {
-		obtained, err := parseMessage([]byte(msg + "\n"))
+		obtained, _, err := parseMessage([]byte(msg + "\n"))
```
rafrombrc: Probably a good idea to capture this value here (and below) and check that it's what we expect.

rafrombrc:
Overall this looks pretty good, thanks for your contribution! In addition to the in-line notes, it would be great if you could add a note to the CHANGES.txt changelog file, as a new feature for the 0.9 release. Cheers!

plugins/statsd/statsd_input.go
wyndhblb: not a real "go" expert, but doing "length" in a few languages can be a performance issue.  if there is not penalty in Go, no there is no need to keep a separate count.

plugins/statsd/statsd_input.go
wyndhblb: okey-dokey

CHANGES.txt
```diff
@@ -74,7 +74,13 @@ Features
 
 * Added ability for HttpListenInput to capture specified HTTP request headers
   and write them as message fields.
+  
+* Added ability for Statsd input handler to treat a malformed stat line 
+  one at time without skipping all the "good" stats in a multi-line input
+  (more like statsd itself).
+  
 
+>>>>>>> upstream/dev
```
bbinet: merge conflict artifact?

PR #1255, opened 2015-01-03
Detect gzip files using the first two magic bytes in the file

ghaering:
(http://en.wikipedia.org/wiki/Gzip#File_format) instead of relying on
the file name ending in .gz.

This makes it possible to use heka with svlogd (part of runit), when
svlogd compresses older log files, as svlogd does not change the
original file name.

### DIFF
--- a/logstreamer/reader.go
+++ b/logstreamer/reader.go
@@ -22,10 +22,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


rafrombrc:
This looks great, and seems to work based on my initial testing. Would you mind adding an entry to the 0.9 features section of the CHANGES.txt file? Thanks.

PR #1259, opened 2015-01-05
add tls support to ES

aaronfeng:
This code is very similar to the way AMQP plugin works.  It checks to see if the server is using https protocol.  If so, TLS is enabled.  Currently, ES does not support TLS on its own, but TLS can be accomplished by proxying it with Apache or Nginx.

There will be another product of ES in the future called the Shield which will support TLS (and other security related stuff).  Presumably this code will also work for people who decide to use Shield.

http://www.elasticsearch.org/overview/shield

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,8 @@
 0.9.0 (2014-MM-DD)
 ==================
 
+* Added TLS support to ElasticSearchOutput (#1259).
+
 Backwards Incompatibilities
 ---------------------------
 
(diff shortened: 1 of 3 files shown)


aaronfeng:
This broken build is not related to my commit (pretty sure).  Removed my commit and same test still fails.

rafrombrc:
Yes, verified that the test is failing even on the versions/0.8 branch, which is very strange. Looking into it now.

PR #1266, opened 2015-01-07
Spin up a separate DecoderRunner for each LogstreamInput instead of

rafrombrc:
forcing all of the streams to share a single one.


plugins/logstreamer/logstreamer_input.go
```diff
@@ -273,6 +275,9 @@ func (li *LogstreamerInput) Run(ir p.InputRunner, h p.PluginHelper) (err error)
 				lsi := NewLogstreamInput(stream, stParser, parserFunc, name,
 					li.hostName)
 				li.plugins[name] = lsi
+				i++
+				dRunner, _ = h.DecoderRunner(li.decoderName, fmt.Sprintf("%s-%s-%d",
```
trink: The dRunner setup should be put in a helper function to avoid code duplication in initialization and rescan

rafrombrc: Done.

Issue #1241, opened 2014-12-22
Heroku log drains decoder


carlanton:
Hello!

Heroku just implemented HTTP(S) support for their log drains [0], i.e. sending log messages from Heroku apps over HTTP. I thought it would be pretty easy to implement a log receiver with Heka, using HttpListenInput and a decoder, but I can't figure out how to write the decoder.

Since Heroku can send multiple log lines in a single request to the HttpListenInput plugin, I need to split them up to multiple messages in the decoder:

```
func (hd *HerokuDecoder) Decode(pack *PipelinePack) (packs []*PipelinePack, err error) {
    lines := strings.Split(*pack.Message.Payload, "\n")
    packs := []*PipelinePack{}
    for _, line := range lines {
        p := // ????
        packs = append(packs, p)
    }
    return packs, nil
}
```

HttpListenInput also sets some other fields for the original pack and I want these to be present in the new packs, so I guess I need to clone them somehow? Maybe all this is easier to do with a sandbox plugin?

Thanks,
Anton

[0] https://devcenter.heroku.com/articles/log-drains#http-s-drains

PR #1267, opened 2015-01-09
[WIP] Add Heroku log input plugin

carlanton:
In #1241 we discussed how to receive log messages from Heroku using HttpListenInput and a decoder or filter.
This pr instead adds a modified version of the HttpListenInput with input validation and log parsing.

With this approach we can:
 - Reject messages (we can require that incoming requests have certain headers etc.)
 - Do the basic Heroku syslog parsing (parsing of the actual message payload is application specific and can be done with sandbox decoder or something)
 - Attach metadata to the messages from the http headers (just DrainToken for now)

Is this something that would be useful to have in Heka core? :octocat:


rafrombrc:
Hrm. This isn't really the direction I'd want to go for the Heka core, no. Instead, if HttpListenInput grew the ability to store arbitrary HTTP headers as fields on the generated message, then all of the rest of the behavior (rejecting messages, pre-parsing, anything else) can be done in the decoder layer. We could provide a Lua module that has Heroku-specific stuff, which could be used as a standalone decoder or in conjunction w/ other code that is doing further parsing, etc.

Duplicating behavior across multiple plugins, blurring the distinction btn the various plugin types responsibility domains, and compiling in vendor specific Go code are all things we want to avoid.

carlanton:
By rejecting messages I meant sending a 500 response code if the body validation/parsing fail, and I don't think that is possible from a decoder. Of course we could just drop the message in Heka and just log that the paring failed. I'll try to give HttpListenInput some love instead!

carlanton:
Any news on the spliter plugin btw?

plugins/heroku/heroku_log_input.go
```diff
@@ -0,0 +1,204 @@
...
+	hostname    string
+}
+
+const LOGPLEX_CONTENT_TYPE = "application/logplex-1"
+
+// Heroku Log Input config struct
+type HerokuLogInputConfig struct {
+	// TCP Address to listen to
+	// Defaults to "0.0.0.0:8325".
+	Address string
```
4r9h: Did you tried to parametrize this module via config? Looks like there's missing `toml:` tag

plugins/heroku/heroku_log_input.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	body, err := ioutil.ReadAll(req.Body)
+	defer req.Body.Close()
+	if err != nil {
+		return fmt.Errorf("Read HTTP request body fail: %s", err.Error())
+	}
+
+	messages := strings.Split(string(body), "\n")
+	acctualMsgCount := len(messages) - 1
```
4r9h: Looks like misspelled variable.

plugins/heroku/heroku_log_input.go
carlanton: @4r9h You don't need the tag thing for "simple" names.

plugins/heroku/heroku_log_input.go
```diff
@@ -0,0 +1,204 @@
...
+
+	return nil
+}
+
+func (hli *HerokuLogInput) Run(ir pipeline.InputRunner, h pipeline.PluginHelper) (err error) {
+	hli.ir = ir
+	hli.hostname = h.Hostname()
+	err = hli.starterFunc(hli)
+	if err != nil {
+		return err
```
4r9h: Missing LogError.

plugins/heroku/heroku_log_input.go
```diff
@@ -0,0 +1,204 @@
...
+
+func (hli *HerokuLogInput) RequestHandler(req *http.Request) error {
+	if req.Method != "POST" || len(req.URL.Path) < 2 ||
+		req.Header.Get("Content-Type") != LOGPLEX_CONTENT_TYPE {
+		return fmt.Errorf("Bad request")
+	}
+
+	msgCount, err := strconv.Atoi(req.Header.Get("Logplex-Msg-Count"))
+	if err != nil {
+		return fmt.Errorf("Could not parse Logplex-Msg-Count: %s", err.Error())
```
4r9h: Personally I think it would be good to create identificators like InvalidMsgCountError to easier handling later.

plugins/heroku/heroku_log_input.go
```diff
@@ -0,0 +1,204 @@
...
+	"time"
+)
+
+type HerokuLogInput struct {
+	conf        *HerokuLogInputConfig
+	listener    net.Listener
+	stopChan    chan bool
+	ir          pipeline.InputRunner
+	server      *http.Server
+	starterFunc func(hli *HerokuLogInput) error
```
4r9h: It would be super to add some docstring with information about purpose of this callback(?).

4r9h:
Maybe it would be good to split repository a little and add some "contrib" submodule with things like this? I mean, i'm not sure if support for heroku should be a part of official maintenance.
@carlanton i did a little review because github didn't display me Rob's comment. I hope you don't mind.

carlanton:
@4r9h Thank you for the review! :+1: I agree with you (and Rob) that this shouldn't be in Heka core. I've made some modifications to the HttpListenInput plugin in a local branch and once we have splitter plugins (see #1241) all parsing can be done in a sandbox decoder instad!

PR #1274, opened 2015-01-14
Reformats logging strings and moves info messages to stdout. Fixes #105

### LINKED ISSUE #105: Use custom logger to distinguish btn stdout and stderr
Currently Heka is just using Go's default logger to generate log output and deliver error messages. This works, but all of the output goes to stderr. We need to also use an alternate logger which will instead send non-error messages to stdout, update all explicit log calls to use the right one, and update the global and plugin runner level LogMessage and LogError methods to use the right one.

4r9h:
@rafrombrc  Hi, its initial version of code to your review and decide on direction of changes. 
Sample output looks like:
[INFO][InputRunner:TcpInput] 2015/01/14 00:55:46.411353 stopped

### DIFF
--- a/pipeline/plugin_runners.go
+++ b/pipeline/plugin_runners.go
@@ -22,6 +22,7 @@ import (
 	"github.com/mozilla-services/heka/client"
 	"github.com/mozilla-services/heka/message"
 	"log"
+	"os"
 	"sync"
 	"time"
 )
@@ -196,6 +197,8 @@ type iRunner struct {
 	sendDecodeFailures bool
 	useMsgBytes        bool
 	deliver            DeliverFunc
+	logInfo            *log.Logger
+	logError           *log.Logger
 }
 
 func (ir *iRunner) Ticker() (ticker <-chan time.Time) {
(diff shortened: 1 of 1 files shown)


rafrombrc:
Sorry to take so long to look at this. Finally getting to it now.

Unfortunately, I think this isn't the right approach. I don't think there's much win to having two different loggers for each type of PluginRunner. The only thing it buys us is the ability to inject type into the logger itself, but that's easily enough just written into the output text in the various LogError / LogMessage methods.

I'd have LogError just use the global logger, as it's already doing. At startup we can create an alternate logger for stdout and store it in the GlobalConfigStruct, and every runner's LogMessage method can use that.

4r9h:
Okay, so i'll stick to the previous version of changes.

On Thu, Jan 15, 2015 at 11:59 PM, Rob Miller <notifications@github.com>
wrote:

> Sorry to take so long to look at this. Finally getting to it now.
>
> Unfortunately, I think this isn't the right approach. I don't think
> there's much win to having two different loggers for each type of
> PluginRunner. The only thing it buys us is the ability to inject type into
> the logger itself, but that's easily enough just written into the output
> text in the various LogError / LogMessage methods.
>
> I'd have LogError just use the global logger, as it's already doing. At
> startup we can create an alternate logger for stdout and store it in the
> GlobalConfigStruct, and every runner's LogMessage method can use that.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mozilla-services/heka/pull/1274#issuecomment-70179174>
> .
>



-- 
jarek@reijutsu:~$ fortune
You have Egyptian flu: you're going to be a mummy.
jarek@reijutsu:~$ fortune
You now have Asian Flu.

rafrombrc:
Superseded by #1282.

PR #1277, opened 2015-01-16
SmtpOutput: encode email subject with base64

liukun:
Email headers must contain only ASCII characters.
http://ncona.com/2011/06/using-utf-8-characters-on-an-e-mail-subject/

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -39,6 +39,8 @@ Bug Handling
 * Hekad now respects --config parameter and loads default config path
   instead of printing help and exit (#1239).
 
+* SmtpOutput now encodes email subject when necessary (#1277).
+
 Features
 --------
 
--- a/plugins/smtp/all_spec_test.go
+++ b/plugins/smtp/all_spec_test.go
@@ -1,0 +1,30 @@
+/***** BEGIN LICENSE BLOCK *****
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+#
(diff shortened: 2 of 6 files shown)


rafrombrc:
Thanks so much for this contribution! Sorry to be slow to respond.

This is a great catch, I definitely want to land it, but I'm thinking it might be better to use quoted-printable instead of base64, since that makes the raw version more human readable. Would you mind changing that?

Also, it would be great if you could add a note to the "bug handling" section for 0.9 in the CHANGES.txt file.

Thanks!

liukun:
@rafrombrc I'd like to change, and is checking Python implementations. Seems not too trivial though, for example, in Python:

    from email.header import Header
    print Header("XXXXXXX - 5 personnes vous ont nommé guide 1234235哈哈", 'utf-8', 76, 'Subject')

gives

    =?utf-8?q?XXXXXXX_-_5_personnes_vous_ont_nomm=C3=A9_guide_123423?=
     =?utf-8?b?NeWTiOWTiA==?=

which uses the shortest encoding and limits line length. I'll try my best to cover these when I have time.

BTW, there seems no built-in quoted-printable encoder in Go? Is it OK to import a new package from GitHub?

rafrombrc:
Oof re: no encoder. Didn't check that, sorry. No, I don't want to add another dependency. We can stick w/ base64 for now.

liukun:
Err, do I need create new pull request to get rid of merge conflicts?

4r9h:
@liukun no, you can resolve conflicts by merging or rebasing and after that you should push changes with --force. It will update PR accordingly.

liukun:
@4r9h thank you! will try that tomorrow.

liukun:
Ready to merge.

PR #1278, opened 2015-01-17
Add disk-based buffering for ElasticSearch output

highlyunavailable:
As discussed on the Heka mailinglist, this is a commit to introduce a
use_buffering mode to the ElasticSearch output plugin. It changes the default
behavior to buffer to disk, like the TCP Output plugin, with all the other
options of the TCP Output plugin and the additional option to set use_buffering
to false to disable it (this just dumps any failed packets to nowhere, though -
no retry logic).

This also buffers in "chunks" rather than individual messages, so there still
is the tiny possibility that if a message has been received and hasn't filled
the commit buffer yet, it can be lost if Heka dies, but the default commit
buffer is sent every 1 second so this state is a lot better than quietly
dropping messages if they fail to send even once.

### DIFF
--- a/plugins/elasticsearch/elasticsearch.go
+++ b/plugins/elasticsearch/elasticsearch.go
@@ -115,10 +115,8 @@ func (o *ElasticSearchOutput) SetName(name string) {
 func (o *ElasticSearchOutput) Init(config interface{}) (err error) {
 	o.conf = config.(*ElasticSearchOutputConfig)
 
-	if !o.conf.UseBuffering {
-		o.batchChan = make(chan []byte)
-		o.backChan = make(chan []byte, 2)
-	}
+	o.batchChan = make(chan []byte)
+	o.backChan = make(chan []byte, 2)
 	var serverUrl *url.URL
 
 	if serverUrl, err = url.Parse(o.conf.Server); err == nil {
@@ -186,9 +184,7 @@ func (o *ElasticSearchOutput) Run(or OutputRunner, h PluginHelper) (err error) {
 		o.bufferedOut.Start(o, outputError, outputExit, stopChan)
 	}
 
-	if !o.conf.UseBuffering {
-		go o.committer(outputExit)
-	}
+	go o.committer(outputExit, &count)
 
 	for ok {
 		select {
@@ -198,15 +194,11 @@ func (o *ElasticSearchOutput) Run(or OutputRunner, h PluginHelper) (err error) {
 			// Closed inChan => we're shutting down, flush data
 			if !ok {
 				if len(outBatch) > 0 {
-					if !o.conf.UseBuffering {
-						o.batchChan <- outBatch
-						outBatch = <-o.backChan
-					}
+					o.batchChan <- outBatch
+					outBatch = <-o.backChan
 					atomic.AddInt64(&o.processMessageCount, int64(count))
 				}
-				if o.batchChan != nil {
-					close(o.batchChan)
-				}
+				close(o.batchChan)
 				stopChan <- true
 				<-outputExit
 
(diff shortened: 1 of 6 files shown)


plugins/elasticsearch/elasticsearch.go
```diff
@@ -166,43 +212,66 @@ func (o *ElasticSearchOutput) receiver(or OutputRunner, wg *sync.WaitGroup) {
 				outBatch = append(outBatch, outBytes...)
 				if count = count + 1; o.bulkIndexer.CheckFlush(count, len(outBatch)) {
```
rafrombrc: Since `count` is now being manipulated from multiple goroutines, we have to start using the synchronized `atomic` API for all access and mutation.

rafrombrc: Which means we'll probably want to define is as int64 instead of just int.

plugins/elasticsearch/elasticsearch.go
```diff
@@ -166,43 +212,66 @@ func (o *ElasticSearchOutput) receiver(or OutputRunner, wg *sync.WaitGroup) {
...
+				}
 				// This will block until the other side is ready to accept
-				// this batch, freeing us to start on the next one.
+				// this batch, so we can't get too far ahead.
 				o.batchChan <- outBatch
 				outBatch = <-o.backChan
-				count = 0
+				atomic.AddInt64(&o.processMessageCount, int64(count))
 			}
+			count = 0
```
rafrombrc: Same re: count.

highlyunavailable:
I'm only ever writing to `count` from the `Run` goroutine, despite the fact that `committer` reads the pointed-to int. Does it still need to be atomic? I've got no problem adding that, I just didn't think it needed to be atomic with 1 writer and multiple readers.

plugins/elasticsearch/elasticsearch.go
highlyunavailable: I'm only ever writing to count from the Run goroutine, despite the fact that committer reads the pointed-to int. Does it still need to be atomic? I've got no problem adding that, I just didn't think it needed to be atomic with 1 writer and multiple readers.

highlyunavailable: I guess so, according to https://golang.org/ref/mem. Good to know.

rafrombrc: Yup, as you discovered, the other goroutines might cache the value they have and not get updates when the mutation happens elsewhere. For higher level data structures there's RWMutex to distinguish btn read locks and write locks, but in this case the atomic API should suffice.

plugins/elasticsearch/elasticsearch.go
```diff
@@ -166,43 +212,66 @@ func (o *ElasticSearchOutput) receiver(or OutputRunner, wg *sync.WaitGroup) {
...
 					}
+					count = 0
 				}
 			}
 		case <-ticker:
 			if len(outBatch) > 0 {
+				if o.conf.UseBuffering {
+					if err = o.bufferedOut.RollQueue(); err != nil {
+						or.LogError(err)
+						return
```
rafrombrc: I know you copied this from the TcpOutput, but I'm realizing that it's a bit off. First, an error returned from Run will automatically get logged, so there's no need to log it inside the Run method. Second, that error will be a low level one returned from an os.OpenFile call, so we should probably decorate it a bit to add some context for the poor folks who have to figure out why Heka just shut down on them. I recommend replacing both of these lines with:

 `return fmt.Errorf("can't create new buffer queue file: %s", err.Error())`

plugins/elasticsearch/elasticsearch.go
```diff
@@ -363,8 +432,48 @@ func (u *UDPBulkIndexer) Index(body []byte) error {
 	return nil
 }
 
+func (o *ElasticSearchOutput) queueFull(buffer []byte, recordCount int) bool {
```
rafrombrc: Would appreciate a doc comment here to explain this function's contract, esp what the returned bool means.

rafrombrc:
Okay, I made my first pass through. Overall, this is solid code, thanks so much for contributing it, and for staying engaged and integrating all of my feedback. It's very much appreciated!

In addition to the small in-line notes I made, there's another thought I have. Right now we're always spinning up a committer goroutine, no matter what. When buffering is in play, though, the committer goroutine isn't strictly necessary, since the buffered output already has its own separate goroutines that isolate the message intake from the outgoing ES requests. I couldn't say whether this has any impact w/o testing, though, and since this is all going to change when the buffering stuff is redone it's probably not worth worrying about too much.

Beyond that, we need to update the docs to reflect the new plugin options, and to add an entry to the features section of the 0.9 notes in CHANGES.txt.

Thanks!

highlyunavailable:
Solid comments, thanks.

Yeah, you're right about the goroutine for flushing not being needed for buffered operations. One other thing I thought of - if I drop the `committer` goroutine to be a dumb memory buffered sender and only use it when not buffering to disk, then move the "processed message count" increment operation to the Run routine, I don't need to worry about making `count` atomic because it won't cross goroutine boundaries then.

rafrombrc:
Good point, maybe it is worth making the change then. I'll leave it up to you.

highlyunavailable:
@rafrombrc Have you gotten a chance to review this? I screwed up the merge a little trying to get rid of the conflicts but I've cleaned up the channel stuff a bit so it only tries to use an outputter channel if we're not in buffered mode.

rafrombrc:
Hi... Sorry for taking so long, I've had my head down trying to wrap up some major changes before I go on vacation for a week in 4 days.

Your changes were pretty good, but my goal was to have either a committer goroutine *or* the BufferedOutput in play, not both. I took the work you had done and tweaked things a bit so this is the case, and cleaned things up just a bit. I also started pushing the batch record count through the batchChan to the committer so we didn't have to worry about touching that variable from separate goroutines.

Anyway, it's merged now, and it seems to be working w/ my superficial local testing. Give it a spin and see how it works. And thanks for contributing!

highlyunavailable:
No problem. Fair enough, I'm not fussed about any changes you made. Thanks a bunch!

PR #1282, opened 2015-01-21
Redirecting logs to stdout and errors to stderr. Fixes #105

### LINKED ISSUE #105: Use custom logger to distinguish btn stdout and stderr

4r9h:
@rafrombrc Hi, its my next attempt at fixing #105. Waiting for your input, always grateful for your reviews.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -19,6 +19,9 @@ Backwards Incompatibilities
   decoding might need to happen in separate goroutines, through `Deliverer`
   objects available from the `NewDeliverer` method.
 
+* Hekad and all clients started using stdio for informational messages and
+  stderr for error messages.
+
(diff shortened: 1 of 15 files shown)


rafrombrc:
This looks great, thanks! Can you add an entry to CHANGES.txt? I think it should go in the "backwards incompatibilities" section since it's a fairly significant operational change.

4r9h:
@rafrombrc Okay, could you look now?

PR #1286, opened 2015-01-22
prepend fields with custom vars

steverweber:
prepend fields with custom vars... might be helpfull... or just a horrible idea.

```
[influxdb]
type = "SandboxEncoder"
filename = "lua_encoders/schema_influx.lua"
    [influxdb.config]
    series = "heka_%{Hostname}"
    prepend_fields = "%{Type}_"
    skip_fields = "Pid EnvVersion FilePath Payload Hostname Type"
```

### DIFF
--- a/sandbox/lua/encoders/schema_influx.lua
+++ b/sandbox/lua/encoders/schema_influx.lua
@@ -41,6 +41,7 @@ Config:
     filename = "lua_encoders/schema_influx.lua"
         [influxdb.config]
         series = "heka.%{Logger}"
+        prepend_fields = "%{Type}_"
         skip_fields = "Pid EnvVersion"
 
     [InfluxOutput]
@@ -62,6 +63,7 @@ Config:
 require "cjson"
 require "string"
 
+local prepend_fields  = read_config("prepend_fields") or ""
 local series_orig  = read_config("series") or "series"
 local series = series_orig
 local use_subs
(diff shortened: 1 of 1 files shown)


steverweber:
the codes kinda ugly... this PR is more food for thought.  feel free to close.

sandbox/lua/encoders/schema_influx.lua
```diff
@@ -139,7 +141,7 @@ function process_message()
 
     local place = 2
     for _, field in ipairs(used_base_fields) do
-        columns[place] = field
+        columns[place] = string.gsub(prepend_fields, "%%{([%w%p]-)}", sub_func)..field
```
rafrombrc: For performance reasons, this string.gsub action should happen only once, before the fields loop, since it will give the same result every time. Also, it should only happen *if* the user specifies a `prepend_fields` value, we shouldn't waste cycles on a gsub call if nothing was specified.

rafrombrc:
I'm fine w/ this feature, but in addition to the performance note I made, we'd want to have the new config option documented in the opening comment (which gets pulled in to the docs). You added it to the example config, but you didn't explicitly list it in the prose part of the docs. Also we'd want to add a note the to "features" section for 0.9 in the CHANGES.txt changelog file.

rafrombrc:
Closing due to lack of interest... ;)

steverweber:
Yap that's fair.  I'll work on something useful tomorrow (that /proc/stat
plugin)

PR #1294, opened 2015-01-29
Add FileOutput date rotation support

### LINKED ISSUE #976: Add FileOutput file rotation
We've had several requests for FileOutput to be able to do file rotation w/o the use of an external rotation tool, in part b/c less tools, and in part b/c Heka needs to get a HUP signal to actually notice that a file has been rotated out from under it, and the person running Heka doesn't always have access to when rotation has happened and when HUP needs to be sent.

bbinet:
(fix #976)

This is my very first golang patch, so please be indulgent ;-)

### DIFF
--- a/plugins/file/file_output.go
+++ b/plugins/file/file_output.go
@@ -31,6 +31,7 @@ import (
 // Output plugin that writes message contents to a file on the file system.
 type FileOutput struct {
 	*FileOutputConfig
+	path       string
 	perm       os.FileMode
 	flushOpAnd bool
 	file       *os.File
@@ -43,10 +44,19 @@ type FileOutput struct {
 // ConfigStruct for FileOutput plugin.
 type FileOutputConfig struct {
 	// Full output file path.
+	// If date rotation is in use, then the output file name can support
+	// Go's time.Format syntax to embed timestamps in the filename:
+	// http://golang.org/pkg/time/#Time.Format
 	Path string
 
 	// Output file permissions (default "644").
 	Perm string
+
+	// Interval at which the output file should be rotated, in hours.
+	// Only the following values are allowed: 0, 1, 4, 12, 24
+	// The files will be named relative to midnight of the day.
+	// (default 0, i.e. disabled). Set to 0 to disable.
+	RotationInterval uint32 `toml:"rotation_interval"`
 
 	// Interval at which accumulated file data should be written to disk, in
 	// milliseconds (default 1000, i.e. 1 second). Set to 0 to disable.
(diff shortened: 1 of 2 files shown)


plugins/file/file_output.go
```diff
@@ -257,11 +296,15 @@ func (o *FileOutput) committer(or OutputRunner, wg *sync.WaitGroup) {
 				// Channel is closed => we're shutting down, exit cleanly.
 				break
 			}
+			if err = o.tryRotateFile(); err != nil {
+				panic(fmt.Sprintf("FileOutput unable to rotate file '%s': %s",
```
4r9h: I would change it into LogError

plugins/file/file_output.go
```diff
@@ -126,14 +151,28 @@ func (o *FileOutput) Init(config interface{}) (err error) {
...
-	o.file, err = os.OpenFile(o.Path, os.O_WRONLY|os.O_APPEND|os.O_CREATE, o.perm)
+	o.file, err = os.OpenFile(o.path, os.O_WRONLY|os.O_APPEND|os.O_CREATE, o.perm)
+	return
+}
+
+func (o *FileOutput) tryRotateFile() (err error) {
+	if o.RotationInterval == 0 {
+		return
+	}
+	t := time.Now().Truncate(time.Duration(o.RotationInterval) * time.Hour)
```
4r9h: Generally, your code looks good for me (as an independent reviewer). But, i was thinking about different approach to the problem caused by performance concerns (however, its only my suggestion). Did you thought about using timer to create separate go function which would be called periodically (in periods defined by user) instead of calling it for every write?
Also it would be crucial to create some tests for this new feature.
However, thats just my thoughts @rafrombrc is more important. Anyway, good job and keep on hacking :)

bbinet: Yes, I already thought of you proposal, but it would introduce a second configuration parameter and as I was not sure about the performance penalty, I decided to not do premature optimization and keep the more simple solution for now.

plugins/file/file_output.go
bbinet: Right, will do.

bbinet: But, I initially followed the same direction as the SIGHUP handling code:
https://github.com/bbinet/heka/blob/fileoutput-rotation/plugins/file/file_output.go#L318

So, do you confirm I should change this to LogError?

rafrombrc: Actually, neither panic nor LogError is ideal here, what *should* happen is that the plugin should exit, returning an error from the Run method. Same is true for the panic that's already in there. Doing this will require restructuring the plugin a bit.

plugins/file/file_output.go
rafrombrc: It would definitely be better to have the file rotation triggered by a timer rather than performing  a calculation on every write, yes.

Issue #1295, opened 2015-01-29
write_message can't overwrite field of different type


hobofan:
While trying to convert some fields from string to int in a lua decoder I noticed that it is not possible to overwrite a field with a value of a different type. 

This is the error being thrown:
```
 decode error: FATAL: process_message() /usr/share/heka/lua_decoders/intify.lua:38: write_message() failed
```

rafrombrc:
This is intended behavior. I've finished and merged your delete fields patch, so this should be resolved.

PR #1296, opened 2015-01-29
delete field values through write_message with nil, fixes #842

### LINKED ISSUE #842: Allow the sandbox write_message() to accept a nil value to delete a field.

hobofan:
Allows deletion of field values in decoders through `write_message("fieldname", nil)`, as suggested in #842 

I implemented this as a workaround for #1295. With this PR a field can be overwritten by a value of different type by deleting it first.

### DIFF
--- a/sandbox/lua/lua_sandbox_interface.c
+++ b/sandbox/lua/lua_sandbox_interface.c
@@ -232,6 +232,7 @@ int write_message(lua_State* lua)
     luaL_argcheck(lua, fi >= 0, 4, "field index must be >= 0");
     int ai = luaL_optinteger(lua, 5, 0);
     luaL_argcheck(lua, ai >= 0, 5, "array index must be >= 0");
+    int has_ai = !lua_isnoneornil(lua, 5); // needed for deletion
 
     int result;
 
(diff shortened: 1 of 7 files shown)


sandbox/lua/lua_sandbox_interface.c
```diff
@@ -232,6 +232,7 @@ int write_message(lua_State* lua)
     luaL_argcheck(lua, fi >= 0, 4, "field index must be >= 0");
     int ai = luaL_optinteger(lua, 5, 0);
     luaL_argcheck(lua, ai >= 0, 5, "array index must be >= 0");
+    int has_ai = luaL_optinteger(lua, 5, -1) != -1; // needed for deletion
```
trink: use !lua_isnoneornil

hobofan: fixed

hobofan:
I just noticed that trying to delete a field that doesn't exist will cause a panic. Will fix that tomorrow.

hobofan:
The panic is fixed and now responds with an appropriate error instead.

sandbox/lua/lua_sandbox_interface.c
```diff
@@ -258,6 +259,10 @@ int write_message(lua_State* lua)
             (char*)value, (char*)rep, fi, ai);
         break;
     }
+    case LUA_TNIL: {
+        result = go_lua_delete_message_field(lsb_get_parent(lsb), (char*)field, fi, ai, has_ai);
+        break;
+    }
     default:
         luaL_error(lua, "write_message() only accepts numeric, string, or boolean field values");
```
trink: update error message

trink:
- ideally deleting a value that doesn't exist would be a no-op and not an error
- CHANGES.txt should be update

PR #1299, opened 2015-01-31
Update handling of EOF in framed protobuf streams

mreid-moz:
Per discussion on IRC, this change should make it easier to deal with the final record in a stream.


pipeline/stream_parser.go
```diff
@@ -262,11 +262,20 @@ func NewMessageProtoParser() (m *MessageProtoParser) {
 
 func (m *MessageProtoParser) Parse(reader io.Reader) (bytesRead int, record []byte, err error) {
 	if m.needData {
-		if bytesRead, err = m.read(reader); err != nil {
+		bytesRead, err = m.read(reader)
+		m.readPos += bytesRead
+		if err == io.EOF {
```
trink: None of this code needs to be duplicated. If you have bytes read and err was not nil and not EOF return otherwise fall through.

mreid-moz:
I've updated the approach here, now the code keeps track explicitly of whether we've reached the end of the stream. We also continue returning any well-formed records until EOF, so you should not need to use the GetRemainingData() function unless you have partial message bytes at the end of the stream and you want to do something with them.

mreid-moz:
See updated PR from the correct branch here:
https://github.com/mozilla-services/heka/pull/1305

PR #1300, opened 2015-01-31
Adding functionality to the schema_influx.lua script to structure

acesaro:
messages as an array of series as opposed to the default which is a
single series with each field as a separate column.  This is a
recommended format for performance reasons in v0.9.0+ as documented here:
http://influxdb.com/docs/v0.8/advanced_topics/schema_design.html#migration-to-0.9.0

### DIFF
--- a/sandbox/lua/encoders/schema_influx.lua
+++ b/sandbox/lua/encoders/schema_influx.lua
@@ -31,6 +31,14 @@ Config:
     schema, any other values will be assumed to refer to a dynamic message
     field.
 
+- multi_series (boolean, optional, default false)
+    Instead of submitting all fields to Influxdb as attributes of a single
+    series, submit a series for each field that sets a "value" attribute
+    to the value of the field.  This also sets the name attribute to the
+    series value with the field name appended to it by a ".".
+    This is the recommended by InfluxDB for v0.9 onwards as it is found
+    to provide better performance when querying and aggregating across
+    multiple series.
 
 *Example Heka Configuration*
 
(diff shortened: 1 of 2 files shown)


sandbox/lua/encoders/schema_influx.lua
```diff
@@ -176,13 +170,85 @@ function process_message()
...
+            local field_name = ""
+            local field_columns = {}
+            local field_points = {}
+
+            -- Instantiate a table in the output table for this iteration
+            output[output_index] = {}
+
+            -- Set the name attribute of this table appropriately
+            if use_subs then
+                series = string.gsub(series_orig, "%%{([%w%p]-)}", sub_func).."."..name
```
rafrombrc: The `string.gsub` call will return the same result each time, in a single `process_message` call. It should happen outside the fields loop.

sandbox/lua/encoders/schema_influx.lua
```diff
@@ -176,13 +170,85 @@ function process_message()
...
+            end
+            field_name = series
```
rafrombrc: If use_subs is false then field_name is going to be the same every time. Is that what we want?

acesaro:
Thanks for the notes!  I'll take these back and update accordingly to then update the PR.

sandbox/lua/encoders/schema_influx.lua
```diff
@@ -176,13 +170,85 @@ function process_message()
...
+    values[1] = read_message("Timestamp") / 1e6
+
+    local place = 2
+    for _, field in ipairs(used_base_fields) do
+        columns[place] = field
+        values[place] = read_message(field)
+        place = place + 1
+    end
+
+    local multi_series = read_config("multi_series")
```
rafrombrc: This should be called outside of `process_message` with the other `read_config` call so it doesn't have to be done for every message.

rafrombrc:
Overall this looks great, thanks! In addition to the in-line comments I made, it'd be great if you could add a note to the "features" section of CHANGES.txt for 0.9. Cheers!

inthecloud247:
curious to know your thoughts here, but should this support message batching? or is that outside of the scope of an encoder.

rafrombrc:
@inthecloud247  I wouldn't usually think of batching as part of an encoder's job. It might be a part of an output's functionality, or it might be done using a filter (i.e. a filter can collect a bunch of incoming messages and periodically emit an aggregate message with the batch). If there were specific reasons why either of those would be difficult, and an encoder resolves those, issues, then maybe, but it smells a bit off to me.

PR #1306, opened 2015-02-04
Three column Health Report for dashboard

SamPenrose:
Third revision for @rafrombrc . There are tricky state issues involved in the collapse/expand of Inputs/Outputs, which tend to drift half-open in response to mysterious bootstrap.js events that do not call my handlers. I have a loop setup that closes erroneously half-opened expanders every 30 seconds. As a bonus, we're saving state in cookies so it will survive refresh and restart. Unfortunately, I haven't been able to eliminate one Output half-expansion on Heka restart. The root of the problem lies somewhere in bootstrap; I can take another run at debugging it if it seems important enough.

Summary:
The Health Report display is "pivoted" from tables vertically listed tables, one for each plugin type, to three columns:
  Inputs | Filters | Outputs
Decoders and Encoders are nested underneath the Inputs/Outputs to which they correspond.


rafrombrc:
This work has moved back to the back-burner, closing for now.

PR #1307, opened 2015-02-04
Splitters

rafrombrc:
Finally.


docs/source/developing/plugin.rst
```diff
@@ -30,37 +30,35 @@ You should be familiar with the :ref:`glossary` terminology before proceeding.
...
+provided by Heka. When Heka starts a plugin, it first creates and configures a
+plugin instance of the appropriate type, then it creates a plugin runner
+instance of the appropriate type, passing in the plugin.
 
 For inputs, filters, and outputs, there's a 1:1 correspondence between
 sections specified in the config file and running plugin instances. This is
-not the case for decoders and encoders, however. Decoder and encoder sections
-register possible configurations, but actual decoder and encoder instances
-aren't created until they are used by input or output plugins.
+not the case for splitters, decoders and encoders, however. Splitter, decoder
```
trink: Splitter should be lower case.

rafrombrc: Why? It's the first word in the sentence.

docs/source/developing/plugin.rst
```diff
@@ -467,55 +675,56 @@ Upon processing a message, a filter plugin can perform any of three tasks:
...
-Once a `PipelinePack` has been obtained, a filter plugin can populate its
-`Message` object. The pack can then be passed along to a specific plugin (or
-plugins) as above. Alternatively, the pack can be injected into the Heka
-message router queue, where it will be checked against all plugin message
-matchers, by passing it to the `FilterRunner.Inject(pack *PipelinePack)`
-method. Note that, again as a precaution against message looping, a plugin
-will not be allowed to inject a message which would get a positive response
-from that plugin's own matcher.
+``PluginHelper.PipelinePack(msgLoopCount int)``. The ``msgloopCount`` value to
+be passed in should be obtained from the ``MsgLoopCount`` value on the pac,
```
trink: typo, pack

docs/source/config/splitters/heka_framing.rst
```diff
@@ -0,0 +1,54 @@
...
+- use_message_bytes (bool, optional):
+	The HekaFramingSplitter is almost always used in concert with an instance
+	of ProtobufDecoder, which expects the protocol buffer message data to be
+	available in the PipelinePack's MsgBytes attribute, so `use_message_bytes`
+	defaults to true.
+
+- skip_authentication (bool, optional):
+	Usually if a HekaFramingSplitter identifies an incorrectly signed message,
+	that message will be silently dropped. In some cases, however, such as
+	when loading a stream of protobuf encoded Heka messages from a file system
```
trink: This is a little strange: since we currently don't write signed messages to disk.  The other thing that is weird is if you authenticate the message you end up having to decode the header a second time.

docs/source/developing/plugin.rst
trink: nm, read it as  however<comma>

rafrombrc: Fair enough. I can reword it to be a bit more clear.

docs/source/developing/plugin.rst
```diff
@@ -346,47 +348,251 @@ needed to perform tasks such as listening on a network connection, making
...
+If you're using SplitBytes, then you'll want to call it each time you have a
+new payload of data to process. It will return the number of bytes
+successfully consumed from the provided slice, and any relevant errors
+occurred while processing. It is up to the calling code to decide what to do
+in error cases, or when all of the data isn't consumed.
+
+If you're using SplitStream, then the SplitStream call will block for as long
+as it is consuming data. When data processing pauses or stops, SplitStream
+will exit and return control back to the input, returning either nil or any
+relevant errors. Typically if nil is returned, you'll want to again call
```
intjonathan: "again call SplitStream again" ?

docs/source/config/splitters/heka_framing.rst
rafrombrc: Yeah, we haven't been writing signed messages to disk, but in the previous implementation the LogstreamerInput was explicitly ignoring the authentication. I put this in so it'd be possible to replicate the prior behaviour.

You're right re: the duplicate header decoding. It's technically possible to cache the header from a successful FindRecord call and reuse it in UnframeRecord, since those calls are made serially from within the same goroutine. We could make that an explicit part of the contract, so folks know it's safe to store struct attributes on their splitter btn the two calls. But signed messages are used so infrequently it didn't seem worth it. I'll happily update if you feel differently, though. :)

docs/source/developing/plugin.rst
rafrombrc: Oh, sorry, that was supposed to be "again call SplitStream again again". ;)

Fixing, thanks.

docs/source/developing/plugin.rst
```diff
@@ -346,47 +348,251 @@ needed to perform tasks such as listening on a network connection, making
...
+        SplitStream(r io.Reader, del Deliverer) error
+        GetRemainingData() (record []byte)
+        GetRecordFromStream(r io.Reader) (int, []byte, error)
+        DeliverRecord(record []byte, del Deliverer)
+        KeepTruncated() bool
+        UseMsgBytes() bool
+        SetPackDecorator(decorator func(*PipelinePack))
+    }
+
+Don't let this scare you, however. SplitterRunner's expose some internal
```
intjonathan: You probably meant 'SplitterRunners'.

PR #1317, opened 2015-02-06
Add a "reset" mechanism for SplitterRunners.

mreid-moz:
In order to be able to reuse a SplitterRunner on a new stream, we need a mechanism to reset the state about the stream.

In order to "tail" a stream that could return EOF, but later have more data to read, we also reset the "reachedEOF" parameter after we return the final record in the stream. That will cause the next subsequent request for a record to re-read the underlying stream. Note that this is *not* sufficient for the "new stream" situation above, where we want to discard any leftover data (as well as state) from the previous stream.

### DIFF
--- a/pipeline/splitter_runner.go
+++ b/pipeline/splitter_runner.go
@@ -40,7 +40,6 @@ type SplitterRunner interface {
 	KeepTruncated() bool
 	UseMsgBytes() bool
 	SetPackDecorator(decorator func(*PipelinePack))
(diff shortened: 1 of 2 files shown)


trink:
We will need to add unit tests and update the SplitterRunner documentation

mreid-moz:
I removed ResetStream() - it's not necessary, as calling GetRemainingData() achieves the same thing.

PR #1320, opened 2015-02-07
add proc stat plugins

steverweber:
add support for proccessing /proc/stat

### DIFF
--- a/sandbox/lua/decoders/linux_procstat.lua
+++ b/sandbox/lua/decoders/linux_procstat.lua
@@ -1,0 +1,111 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[[
+
+Parses a payload containing the contents of file `/proc/stat`. This also
+tries to obtain the TickerInterval of the input it recieved the data from, by
+extracting it from a message field named `TickerInterval`.
+
+Config:
+
+- payload_keep (bool, optional, default false)
+    Always preserve the original log line in the message payload.
+
+*Example Heka Configuration*
+
+.. code-block:: ini
+
+    [ProcStats]
+    type = "FilePollingInput"
+    ticker_interval = 1
+    file_path = "/proc/stat"
+    decoder = "ProcStatDecoder"
+
+    [ProcStatDecoder]
+    type = "SandboxDecoder"
+    filename = "lua_decoders/linux_procstat.lua"
+
+*Example Heka Message*
+
+:Timestamp: 2014-12-10 22:38:24 +0000 UTC
+:Type: stats.proc
+:Hostname: yourhost.net
+:Pid: 0
+:Uuid: d2546942-7c36-4042-ad2e-f6bfdac11cdb
+:Logger:
+:Payload:
+:EnvVersion:
+:Severity: 7
+:Fields:
+    | name:"cpu" type:double value:[14384,125,3330,946000,333,0,356,0,0,0]
+    | name:"cpu[1-#]" type:double value:[14384,125,3330,946000,333,0,356,0,0,0]
+    | name:"TickerInterval" type:double value:1
+    | name:"ctxt" type:double value:2808304
+    | name:"btime" type:double value:1423004780
+    | name:"intr" type:double value:[14384,125,3330,0,0,0,0,0,0,0...0]
+    | name:"processes" type:double value:3811
+    | name:"procs_running" type:double value:1
+    | name:"procs_blocked" type:double value:0
+    | name:"softirq" type:double value:[288977,23,101952,19,13046,19217,7,...]
+    | name:"index_count" type:number value:10
+    | name:"index" type:string value:[cpu,cpu#,TickerInterval,ctxt,btime...]
+
+    Cpu fields:
+        1    2    3      4    5        6     7         8       9       10
+        user nice system idle [iowait] [irq] [softirq] [steal] [guest] [guestnice]
+        Note: systems provide user, nice, system, idle. Other fields depend on kernel.
+
+    intr
+        This line shows counts of interrupts serviced since
+        boot time, for each of the possible system interrupts.
+        The first column is the total of all interrupts
+        serviced including unnumbered architecture specific
+        interrupts; each subsequent column is the total for
+        that particular numbered interrupt.  Unnumbered
+        interrupts are not shown, only summed into the total.
+
+--]]
+
+local l = require 'lpeg'
+l.locale(l)
+
+local sp = l.P" "^1
+local num = l.digit^1 / tonumber
+local key = l.C((l.alnum + "_")^1)
+local array = l.Ct((num * sp^0)^1)
+local line =  l.Cg(key * sp * array) * l.P"\n"^0
+
+local grammar = l.Cf(l.Ct("") * line^1, rawset)
+
+local payload_keep = read_config("payload_keep")
+
+local msg = {
+    Type = "stats.procstat",
+    Payload = nil,
+    Fields = nil,
+}
+
+function process_message()
+    local data = read_message("Payload")
+    msg.Fields = grammar:match(data)
+    if not msg.Fields then
+        return -1
+    end
+    if payload_keep then
+        msg.Payload = data
+    end
+    -- var index is a list of all fields found in the payload
+    local count = 1
+    local index = {}
+    for k in pairs(msg.Fields) do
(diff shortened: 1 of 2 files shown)


steverweber:
let me know what else needs fixing.
O also It would be nice if I could use somthing more like regex for the whitelist_pattern... any ideas?... i want to pass it in as a string.. im not sure if lpeg has any support for stuff like that.

steverweber:
while I still have time to hack on this.
feedback please.

trink:
The latest dev branch includes the LPeg 're' module so passing in a string pattern is now possible.

steverweber:
thanks @trink

steverweber:
@bbangert forgot to include you in this PR. Perhaps you have some feedback and could preform some testing.

sandbox/lua/decoders/linux_procstat.lua
```diff
@@ -0,0 +1,109 @@
...
+local l = require 'lpeg'
+l.locale(l)
+
+local sp = l.P" "^1
+local num = l.digit^1 / tonumber
+local key = l.C((l.alnum + "_")^1)
+local array = l.Ct((num * sp^0)^1)
+local line =  l.Cg(key * sp * array) * l.P"\n"^0
+
+local grammar = lpeg.Cf(lpeg.Ct("") * line^1, rawset)
```
trink: Use l.Cf/Ct instead to be consistent with the other grammars.

sandbox/lua/decoders/linux_procstat.lua
```diff
@@ -0,0 +1,109 @@
...
+    type = "FilePollingInput"
+    ticker_interval = 1
+    file_path = "/proc/stat"
+    decoder = "ProcStatDecoder"
+
+    [ProcStatDecoder]
+    type = "SandboxDecoder"
+    filename = "lua_decoders/linux_procstat.lua"
+
+*Example Heka Message*
```
trink: Sample is not representative of the the actual output (missing index entries).

sandbox/lua/decoders/linux_procstat.lua
```diff
@@ -0,0 +1,109 @@
...
+        return -1
+    end
+    if payload_keep then
+        msg.Payload = data
+    end
+    local count = 0
+    local index = {}
+    for k in pairs(msg.Fields) do
+        index[count] = k
+        count = count + 1
```
trink: Lua array counts traditionally start at 1

sandbox/lua/filters/procstat.lua
```diff
@@ -0,0 +1,200 @@
...
+        Type = "procstat",
+        Payload = read_message("Payload"), 
+        Fields = {},
+    }
+    -- read values found in the fields_index into a usable table
+    local wl = whitelist_pattern
+    local data = {}
+    local index_count = read_message("Fields[index_count]")
+    if index_count == nil then return 0 end
+    for i=1, index_count - 1 do
```
trink: Off by 1

sandbox/lua/filters/procstat.lua
```diff
@@ -0,0 +1,200 @@
...
+
+softirq 288977 23 101952 19 13046 19217 7 19125 92077 389 43122
+    Time spent servicing soft-interrupts.
+--]]
+
+-- require "cjson"
+
+
+local l = require 'lpeg'
+require "string"
```
trink: As a matter of style, you should use always the same kind of quotes (single or double) in a program, unless the string itself has quotes; then you use the other quote, or escape those quotes with backslashes.

sandbox/lua/filters/procstat.lua
```diff
@@ -0,0 +1,200 @@
...
+}
+
+local preserve_data = read_config("preserve_data")
+local blacklist_regex = read_config("blacklist_regex") or 'cpu$'
+
+function process_delta(new, old)
+    -- Recursion - check diff in table or value
+    if type(new) ~= type(old) then return nil end
+    if type(old) ~= "table" then
+        return new - old
```
trink: Assumes the fields are numeric or convertible to a numeric bad data could crash this plugin.

sandbox/lua/filters/procstat.lua
```diff
@@ -0,0 +1,200 @@
...
+        Payload = read_message("Payload"), 
+        Fields = {},
+    }
+    -- read values found in the fields_index into a usable table
+    local wl = whitelist_pattern
+    local data = {}
+    local index_count = read_message("Fields[index_count]")
+    if index_count == nil then return 0 end
+    for i=1, index_count - 1 do
+        f = read_message("Fields[index]", 0, i-1)
```
trink: Not sure why the list of fields is being include when they could just be iterated with read_next.  This would simplify a bunch of the following code and also reduce the message size.

sandbox/lua/filters/procstat.lua
```diff
@@ -0,0 +1,200 @@
...
+        end
+    end
+    if proc_stat_previous == nil then
+        proc_stat_previous = data
+        return 0
+    end
+    local delta = process_delta(data, proc_stat_previous)
+    msg.Fields = process_fields(delta)
+    
+    --if preserve_data then
```
trink: remove

steverweber:
I got some time.. @trink ill make them changes and test
Thanks!

steverweber:
I never tested require "re" yet because i'm using the older binary build of heka.

steverweber:
this merge should be fine to merge after the blacklist_regex uses 're'
however some feedback would be nice on the above items.

thanks.

sandbox/lua/filters/procstat.lua
```diff
@@ -0,0 +1,191 @@
...
+
+local proc_stat_previous = nil
+local proc_stat_mappings = {
+    cpu = {
+        'user', 'nice', 'system', 'idle', 'iowait', 'irq',
+        'softirq', 'steal', 'guest', 'guestnice'
+    }
+}
+
+local preserve_data = read_config('preserve_data')
```
trink: Not used

sandbox/lua/filters/procstat.lua
```diff
@@ -0,0 +1,191 @@
...
+
+
+function process_fields(delta)
+    -- proccess deltas and calculate cpu average usage on fields of interest
+    local fields = {}
+    for k, v in pairs(delta) do
+        --if type(v) ~= 'table' then
+        --    msg.Fields[k..'_delta'] = v
+        --end
+        if k:match('cpu') ~= nil then
```
trink: Anchor the match

sandbox/lua/filters/procstat.lua
```diff
@@ -0,0 +1,191 @@
...
+    local stats = {}
+    local msg = {
+        Type = 'procstat',
+        Payload = read_message('Payload'), 
+        Fields = {},
+    }
+    -- read values found in the fields_index into a usable table
+    local wl = whitelist_regex
+    local data = {}
+    local index_count = read_message('Fields[index_count]')
```
trink: You can git rid of index_count and the index Fields in the decoder and you can replace the loop with something like this
```lua
while true do
    local typ, name, value, rep, count = read_next_field()
    if not typ then break end

    if not wl or name:match(wl) then
        if count > 1 then
            local field_name = string.format("Fields[%s]", name)
            data[name][1] = value
            for i=1, count - 1 do
                data[name][i+1] = read_message(field_name, 0, i)
            end
        else
            data[name] = value
        end
    end
end
```

When https://github.com/mozilla-services/heka/issues/1342 is fixed you will be able to pull down the entire message at once and avoid all these read calls.

steverweber: I like this... while i made that ugly for loop i was thinking most of this junk should be done at the sandbox message API.  will this api feature be added in the next release?

sandbox/lua/filters/procstat.lua
steverweber: want to use regex... but yap ill ^cpu it.

sandbox/lua/filters/procstat.lua
steverweber: i think i had it heare as a note because the previous developer was saving or passing the Payload on...
I think ill drop it.

sandbox/lua/decoders/linux_procstat.lua
steverweber: i added the index entries (index_count, index) ... but thses will be removed i hope.

sandbox/lua/filters/procstat.lua
trink: The code above you can use now (that API has been available for a long time).  No ETA on the much nicer way yet.

steverweber:
see new PR.

PR #1322, opened 2015-02-08
Enable AMQP input to use a read-only user

### LINKED ISSUE #1314: AMQP input binding to readonly exchange
I'm trying to set up an AMQP input to an exchange that I only have read permissions to.  It seems the default behaviour is to declare an exchange, which requires readwrite permissions.  It is possible to simply bind to the exchange, if so, its not obvious how to do this from the documentation.

This is my current configuration:
```
[AMQPInput]
url = "amqp://user:password@amqp.company.com/VHOST"
exchange = "MyExchange"
exchange_type = "topic"
exchange_durability = true
exchange_auto_delete =  false
routing_key = "#"
queue = "test"
decoder = "json_decoder"
```

Thanks
Alex

mattrco:
This adds a configuration option, ReadOnly, to AMQP input to
allow users which only have read access to connect to a queue.

The exchange, queue, and binding must be declared prior to
starting heka with a read-only user.

The config option defaults to false. Fixes #1314.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -105,6 +105,8 @@ Features
 
 * Added `cert_path` setting for TLS support to DockerLogInput. 
 
+* Added `read_only` setting to AMQPInput to support read-only users.
+
 0.8.3 (2015-01-08)
 ==================
 
--- a/docs/source/config/inputs/amqp.rst
+++ b/docs/source/config/inputs/amqp.rst
@@ -57,6 +57,12 @@ Config:
     SSL/TLS encryption. This will only have any impact if `URL` uses the
     `AMQPS` URI scheme. See :ref:`tls`.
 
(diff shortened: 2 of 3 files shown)


mattrco:
(Happy to update docs for this when/if it's merged)

ioc32:
Cannot comment on the code (go is not my strong suit) but the feature is very nice. It allows us to exert tighter control over the data pipeline.

Looking forward to see progress here! I can give a hand too if you need the docs updated.

rafrombrc:
This looks fine, but the docs will need to be added *before* we can merge it to dev. Also we'll want to add an entry to the "bug handling" section for 0.9 in the CHANGES.txt changelog file. Thanks!

mattrco:
Thanks for the review @rafrombrc, I've added docs.

Issue #918, opened 2014-06-20
Reimplement reporting infrastructure


rafrombrc:
Right now every plugin can generate custom report data to be included in the dashboard, but there's no consistency or shared reporting infrastructure. This should be revisited to provide at least the following features:
- Message count, error count, and appropriately sampled processing duration values should be provided "for free" without the need for each plugin to implement this independently.
- MultiDecoders should be able to delegate to sub-decoders to extract reporting information from them while still being able to keep track of which data came from which sub-decoder.

PR #1323, opened 2015-02-09
Add a ReportMsg interface to the SandboxOutput

trink:


sandbox/plugins/sandbox_output.go
```diff
@@ -136,7 +137,6 @@ func (s *SandboxOutput) Run(or pipeline.OutputRunner, h pipeline.PluginHelper) (
 				}
 			} else {
 				err = fmt.Errorf("FATAL: %s", s.sb.LastError())
-				or.LogError(err)
```
mreid-moz: Why not log this fatal error anymore?

trink: The error message will be output twice, which is just annoying.

sandbox/plugins/sandbox_output.go
```diff
@@ -171,6 +170,42 @@ func (s *SandboxOutput) Run(or pipeline.OutputRunner, h pipeline.PluginHelper) (
...
+	if s.processMessageSamples > 0 {
+		tmp = s.processMessageDuration / s.processMessageSamples
+	}
+	message.NewInt64Field(msg, "ProcessMessageAvgDuration", tmp, "ns")
+
+	tmp = 0
+	if s.timerEventSamples > 0 {
+		tmp = s.timerEventDuration / s.timerEventSamples
+	}
+	message.NewInt64Field(msg, "TimerEventAvgDuration", tmp, "ns")
```
mreid-moz: Consider introducing an "avg" utility function to save duplicate code.

mreid-moz:
Looks good to me. Up to you whether it's worthwhile adding a function to compute average (and avoid div by 0)

sandbox/plugins/sandbox_output.go
trink: I will leave it like the others for now... but noting it for https://github.com/mozilla-services/heka/issues/918

PR #1325, opened 2015-02-10
Fix Kafka behavior with manual offset when checkpoint file doesn't exist

whd:
An offset of 0 is not always valid, so use oldest instead.

### DIFF
--- a/plugins/kafka/kafka_input.go
+++ b/plugins/kafka/kafka_input.go
@@ -170,9 +170,7 @@ func (k *KafkaInput) Init(config interface{}) (err error) {
 			if err = os.MkdirAll(filepath.Dir(k.checkpointFilename), 0766); err != nil {
 				return
 			}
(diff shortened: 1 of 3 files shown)


plugins/kafka/kafka_input.go
```diff
@@ -170,9 +170,7 @@ func (k *KafkaInput) Init(config interface{}) (err error) {
 			if err = os.MkdirAll(filepath.Dir(k.checkpointFilename), 0766); err != nil {
 				return
 			}
-			if err = k.writeCheckpoint(0); err != nil {
-				return
-			}
+			k.consumerConfig.OffsetMethod = sarama.OffsetMethodOldest
```
trink: We want to start from the oldest point but only the first time we don't have a checkpoint.  Once we start from there we want to keep track of where we are.  This change will always process the full history every time Heka is restarted.

trink:
Please add a note to the CHANGES.txt Bug Handling section.

PR #1328, opened 2015-02-10
Add new message fields.

mreid-moz:
Add message fields for RemoteAddr, URL Path, EnvVersion,
Hostname (the name of the server that handled the request), and
Host (the hostname to which the client sent the request).


plugins/http/http_listen_input.go
```diff
@@ -71,43 +72,56 @@ func defaultStarter(hli *HttpListenInput) (err error) {
...
 func (hli *HttpListenInput) makePackDecorator(req *http.Request) func(*PipelinePack) {
 	packDecorator := func(pack *PipelinePack) {
 		pack.Message.SetType("heka.httpdata.request")
 		pack.Message.SetPid(hli.hekaPid)
 		pack.Message.SetSeverity(int32(6))
-		if field, err := message.NewField("Protocol", req.Proto, ""); err == nil {
+
+		// Host on which heka is running.
+		pack.Message.SetHostname(hli.hostname)
+		pack.Message.SetEnvVersion("1")
```
mreid-moz: I'm not sure if it makes sense to arbitrarily set this to "1", but it seemed preferable to leaving it blank.

trink: Blank takes up no space on the wire but I have no problem with bumping the version when we change the schema.

trink:
Please update the CHANGES.txt

mreid-moz:
CHANGES.txt updated.

mreid-moz:
I updated the `RemoteAddr` and `Host` fields to remove the port info, since it doesn't seem useful, and could be cumbersome for users of these fields to have to parse out the IP.

Issue #1311, opened 2015-02-05
RPM build is failing on the rpath in the new shared libs


trink:
```
CPack: Create package
CPackRPM: Will use GENERATED spec file: /home/build/data-pipeline/build/heka/build/_CPack_Packages/Linux/RPM/SPECS/heka.spec
CPackRPM:Debug: You may consult rpmbuild logs in:
CPackRPM:Debug:    - /home/build/data-pipeline/build/heka/build/_CPack_Packages/Linux/RPM/rpmbuild.err
CPackRPM:Debug: *** + umask 022
+ cd /home/build/data-pipeline/build/heka/build/_CPack_Packages/Linux/RPM/BUILD
+ mv /home/build/data-pipeline/build/heka/build/_CPack_Packages/Linux/RPM/heka-0_9_0-linux-amd64 /home/build/data-pipeline/build/heka/build/_CPack_Packages/Linux/RPM/tmpBBroot
+ exit 0
+ umask 022
+ cd /home/build/data-pipeline/build/heka/build/_CPack_Packages/Linux/RPM/BUILD
+ '[' /home/build/data-pipeline/build/heka/build/_CPack_Packages/Linux/RPM/heka-0_9_0-linux-amd64 '!=' / ']'
+ rm -rf /home/build/data-pipeline/build/heka/build/_CPack_Packages/Linux/RPM/heka-0_9_0-linux-amd64
++ dirname /home/build/data-pipeline/build/heka/build/_CPack_Packages/Linux/RPM/heka-0_9_0-linux-amd64
+ mkdir -p /home/build/data-pipeline/build/heka/build/_CPack_Packages/Linux/RPM
+ mkdir /home/build/data-pipeline/build/heka/build/_CPack_Packages/Linux/RPM/heka-0_9_0-linux-amd64
+ '[' -e /home/build/data-pipeline/build/heka/build/_CPack_Packages/Linux/RPM/heka-0_9_0-linux-amd64 ']'
+ rm -rf /home/build/data-pipeline/build/heka/build/_CPack_Packages/Linux/RPM/heka-0_9_0-linux-amd64
+ mv /home/build/data-pipeline/build/heka/build/_CPack_Packages/Linux/RPM/tmpBBroot /home/build/data-pipeline/build/heka/build/_CPack_Packages/Linux/RPM/heka-0_9_0-linux-amd64
+ /usr/lib/rpm/check-rpaths
*******************************************************************************
*
[...]

PR #1330, opened 2015-02-10
Issue #1311 fix the Centos7 RPM

trink:


whd:
:+1: I've confirmed the RPM builds and installs correctly on CentOS 7 (when built with CMake 3).

PR #1331, opened 2015-02-11
plugin/amqp: expanded field coverage

logicalparadox:
This PR contains a few tweaks I have made to the AMQP plugin to expand it's data coverage. It is functional, but not data-complete; AMQP headers are not (yet) included as heka field(s). This is something I need, but I would like it to be implemented in idiomatic heka so I am open to suggestions on the best way to approach this.

#### Change Summary:

* `plugins/amqp/type.go`: Expose dialer and hub to allow for further custom AMQP plugins based on same connection methodology.
* `plugins/amqp/input.go`: Include most AMQP meta data as Heka message field.

### DIFF
--- a/plugins/amqp/amqp_input.go
+++ b/plugins/amqp/amqp_input.go
@@ -108,7 +108,7 @@ func (ai *AMQPInput) Init(config interface{}) (err error) {
 	}
 
 	if ai.amqpHub == nil {
-		ai.amqpHub = getAmqpHub()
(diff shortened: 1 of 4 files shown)


logicalparadox:
So this clearly is nowhere near compatible with the new Splitter approach. I will take the delivered fields feature back to drawing board but expect a new PR shortly for the exposed dialer functionality.

PR #1337, opened 2015-02-13
Set Type on Tcp and Udp Input's

hynd:
Since the Splitter code landed, the "Type" field is no longer getting set by the TcpInput and UdpInput.
If it wasn't intentional, please consider this PR to reinstate it :)
(I changed the value of the Type slightly to match what some of the others looked like - ie; lowercase, prefixed with 'heka.')

Also tweaked the formatting for the startup messages to calm my OCD.

### DIFF
--- a/pipeline/pipeline_runner.go
+++ b/pipeline/pipeline_runner.go
@@ -215,7 +215,7 @@ func Run(config *PipelineConfig) {
 			outputsWg.Done()
 			continue
 		}
-		LogInfo.Println("Output started: ", name)
+		LogInfo.Println("Output started:", name)
 	}
 
 	for name, filter := range config.FilterRunners {
@@ -225,7 +225,7 @@ func Run(config *PipelineConfig) {
 			config.filtersWg.Done()
 			continue
 		}
-		LogInfo.Println("Filter started: ", name)
+		LogInfo.Println("Filter started:", name)
 	}
 
 	// Finish initializing the router's matchers.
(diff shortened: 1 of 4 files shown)


pipeline/pipeline_runner.go
```diff
@@ -215,7 +215,7 @@ func Run(config *PipelineConfig) {
 			outputsWg.Done()
 			continue
 		}
-		LogInfo.Println("Output started: ", name)
+		LogInfo.Println("Output started:", name)
```
trink: Why the removal of the space? Looking at the other error messages a colon space seems to be more of the convention.

trink:
The change in naming should also be noted in CHANGES.txt.  I will leave it for @rafrombrc to merge since I am unsure if he intentionally removed Type to begin with.

pipeline/pipeline_runner.go
hynd: Println adds a space between each operand anyways - they were getting two spaces after the colon, which is what made them stand out against all the other startup messages.

trink: +1

trink:
@rafrombrc we should audit the `heka.*` log message Types again; I know I threw in a heka.kafka default which is a probably a bad idea if they are suppose to represent local messages only.

rafrombrc:
I knew I was leaving out the type setting code. Having all of the messages of type "NetworkInput" didn't seem very useful to me. I feel similarly about `heka.tcpinput` and `heka.udpinput`; there's not really much value to it. Still, it probably does make sense to set the message type to *something*... I'm thinking using the name of the input is probably the best bet. I'll make this change, update the changelog, and merge.

PR #1338, opened 2015-02-13
Make rsyslog decoder's hostname setting optional

ioc32:
Under some circumstances the hostnames found in syslogs are not canonical. This might cause trouble because logs from worker30.cluster1.example.net and worker30.cluster2.example.net will both show up as worker30.

In order to solve this, we are using Logstreamer's 'hostname' config option to set msg.Hostname to the fully qualified domain names we expect.

This PR adds the 'hostname_keep' boolean config option to the rsyslog decoder so we can control if we want it overridden or not. Its defaults do not change the current behavior of the rsyslog decoder.

I can update the PR with the required changes to the documentation, should you be interested in pulling the commits in. As I said, this change works well for us. Feel free to implement the suggestion or drop it :-)


sandbox/lua/decoders/rsyslog.lua
```diff
@@ -50,6 +52,7 @@ local syslog = require "syslog"
 
 local template = read_config("template")
 local msg_type = read_config("type")
+local hostname_keep = read_config("hostname_keep") or false
```
trink: Don't need the `or false`

ioc32: Ugh! Thanks!

sandbox/lua/decoders/rsyslog.lua
```diff
@@ -93,14 +96,16 @@ function process_message ()
...
+        fields.hostname = nil
+        fields.source = nil
+    end
 
     msg.Payload = fields.msg
     fields.msg = nil
 
     msg.Fields = fields
-    inject_message(msg)
+    if not pcall(inject_message, msg) then return -1 end
```
trink: I assume this was added to catch an output_limit error since the grammar ensures the message is well formed.  So we are changing the behaviour from termination the dropping the data... I am good with that, any objections?

CHANGES.txt should be updated

ioc32: Not at all. PR updated!

PR #1345, opened 2015-02-19
Splitter buffer size

rafrombrc:


trink:
I was surprised to see that read will fill a large buffer in one call (i.e. 50MB) but it appears to work.

One thing that I don't like about the design is that one always has to set buffer_size to the maximum size instead being able to grow it **if** needed (like the old implementation) but if @mreid-moz is happy with it then r+

trink:
My initial test was with Logstreamer input.  I retested with HttpListenInput for @mreid-moz and the body is split into multiple reads even on something as small as a 35KiB payload. So this will not work.

PR #1365, opened 2015-02-25
Cbuf influx encoder

rafrombrc:


trink:
The cbuf utils are useful in the sandbox and not just Heka.  There is also a existing module for cbufd in the sandbox at some point in time these should be merged (the grammar implementation is much more succinct)

sandbox/lua/modules/cbuf_utils.lua
```diff
@@ -0,0 +1,151 @@
...
+end
+
+-- [[ Public Interface --]]
+
+function parse_cbuf(cbuf_text)
+    local rows = {}
+    local headersText, headers, ok
+    local first = true
+    for row in string.gmatch(cbuf_text, "[^\n]+") do
+        if first == true then
```
trink: You need to account for the possible dygraph options header too.

sandbox/lua/modules/cbuf_utils.lua
```diff
@@ -0,0 +1,151 @@
...
+    end
+    if headers.columns ~= #headers.column_info then
+        return
+    end
+    return true
+end
+
+-- [[ Public Interface --]]
+
+function parse_cbuf(cbuf_text)
```
trink: See https://github.com/mozilla-services/lua_sandbox/blob/dev/modules/cbufd.lua#L47 for a similar parser

sandbox/lua/modules/cbuf_utils.lua
```diff
@@ -0,0 +1,151 @@
...
+                first = false
+            end
+        else
+            -- Parse the row values
+            local values = {}
+            for value in string.gmatch(row, "[^\t]+") do
+                if value ~= "nan" then
+                    value = tonumber(value)
+                    if not value then
+                        value = "nan"
```
trink: Why not actually set it to an NaN value?

sandbox/lua/modules/cbuf_utils.lua
```diff
@@ -0,0 +1,151 @@
...
+                end
+                values[#values+1] = value
+
+            end
+            rows[#rows+1] = values
+        end
+    end
+    -- Last sanity check that headers match the actual row count.
+    local num_rows = #rows
+    if num_rows < 3 or num_rows ~= headers.rows then
```
trink: 2 rows are allowed

sandbox/lua/modules/cbuf_utils.lua
rafrombrc: So it would be consistent w/ the "nan" strings that are already in the data structure. Moot if we replace it w/ the grammar, though.

sandbox/lua/modules/cbuf_utils.lua
```diff
@@ -0,0 +1,151 @@
...
+    end
+    local rows_elapsed = elapsed_time / headers.seconds_per_row
+
+    if rows_elapsed >= headers.rows - 1 then
+        -- All the rows we've consumed so far have been advanced, start at the
+        -- beginning.
+        return 1
+    end
+    -- We've partially advanced, return the correctly computed starting row.
+    return headers.rows - rows_elapsed
```
trink: So data that may have change in rows before the last_time in lost?

sandbox/lua/modules/cbuf_utils.lua
trink: if someone was using this parser and round trip the data back into a cbuf it would not work.
```lua
value = "nan"
cb:set(ts, 1, value)
```

trink:
In general combining circular buffers this way is a very bad idea (almost always leading to some data loss).  I would recommend updating these plugins to use the delta output, it is the reason it was introduced.

trink:
Removed my comments: 
Sorry, I just jumped right into implementation review mode but taking a step back I think we should change the entire approach (as per https://github.com/mozilla-services/heka/pull/1365#issuecomment-76038929) which makes the other comments moot.

rafrombrc:
Closing until revisiting w/ different approach.

PR #1374, opened 2015-02-27
Add configurable max message size to heka-cat

mreid-moz:
This lets heka-cat work with messages larger than the default 64k. A 4MB maximum is probably overkill, but it should have minimal impact as long as the actual messages encountered are small.


cmd/heka-cat/main.go
```diff
@@ -58,6 +59,14 @@ func main() {
 		os.Exit(1)
 	}
 
+	if *flagMaxMessageSize < uint64(4294967295) {
```
trink: math.MaxUint32

PR #1376, opened 2015-02-27
Token every nth

rafrombrc:


plugins/http/http_listen_input.go
```diff
@@ -184,6 +184,11 @@ func (hli *HttpListenInput) RequestHandler(w http.ResponseWriter, req *http.Requ
 			} else {
 				sRunner.DeliverRecord(record, nil)
 			}
+		} else if len(record) == 0 && sRunner.IncompleteFinal() && err == io.EOF {
```
trink: Hmm, different constructs between the two http plugins.  In this case `len(record) == 0` is not needed (if it had a length the deliver flag would have also been true).

rafrombrc:
Okay, merged the two HTTP related stream splitting implementations into one.

PR #1396, opened 2015-03-06
Dashboard: Close plugin list file

mattbostock:
Close the file once we've finished writing to it.


trink:
It is automatically closed when it goes out of scope, so this is unneeded.

PR #1400, opened 2015-03-08
Added a config flag to replace all dots in stats filters

lyrixx:
Useful when this filter runs with graphite.
Because graphite use a dot as a namespace separator.
And so, if a Type for exmeple contains a dot, it will
mess up the graphite storage.

note: I'm a real beginner with golang. I'm sure the code could be better ;)

### DIFF
--- a/plugins/statsd/stat_filter.go
+++ b/plugins/statsd/stat_filter.go
@@ -21,6 +21,7 @@ import (
 	"github.com/mozilla-services/heka/message"
 	. "github.com/mozilla-services/heka/pipeline"
 	"strconv"
+	"strings"
 )
 
 // Simple struct representing a single statsd-style metric value.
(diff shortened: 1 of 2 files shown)


plugins/statsd/stat_filter.go
```diff
@@ -107,7 +109,16 @@ func (s *StatFilter) Run(fr FilterRunner, h PluginHelper) (err error) {
 
 		// We matched, generate appropriate metrics
 		for _, met := range s.metrics {
+			if met.ReplaceDot {
+				var valuesTmp = make(map[string]string)
```
rafrombrc: `valuesTmp` isn't needed, you can mutate the `values` map directly in the loop.

lyrixx: Thanks, fixed.

Issue #1405, opened 2015-03-10
read_message("raw") can cause a panic if MsgBytes is empty.


mreid-moz:
I attempted to use read_message("raw") in a SandboxEncoder, and got an error of "panic: runtime error: index out of range".  The stack trace pointed to [the sandbox code](https://github.com/mozilla-services/heka/blob/dev/sandbox/lua/lua_sandbox.go.in#L325).

Reading the message fields rather than "raw" worked as expected.

PR #1407, opened 2015-03-11
Issue #1405 Invalid index reading 'raw' message

trink:


pipeline/pipeline_runner.go
```diff
@@ -158,7 +158,7 @@ type PipelinePack struct {
 // Returns a new PipelinePack pointer that will recycle itself onto the
 // provided channel when a message has completed processing.
 func NewPipelinePack(recycleChan chan *PipelinePack) (pack *PipelinePack) {
-	msgBytes := make([]byte, message.MAX_MESSAGE_SIZE)
+	msgBytes := make([]byte, 0, message.MAX_MESSAGE_SIZE)
```
rafrombrc: This worries me a bit. `pack.Zero()` sets MsgBytes length to the full capacity, and our code generally copies into MsgBytes and then resizes down, rather than resizing up and then copying. I'm okay w/ changing this, but I'd rather do it in a separate PR, since it requires digging into all of the places where we set pack.MsgBytes to make sure we don't silently truncate b/c the length is too short.

trink: "our code generally copies into MsgBytes and then resizes down" Can you point me to where you are referring, The length is meaningless when populating it with something new and we should only be dealing with capacity (like this):

```Go
		messageLen := len(unframed)
		if messageLen > cap(pack.MsgBytes) {
			pack.MsgBytes = make([]byte, messageLen)
		}
		pack.MsgBytes = pack.MsgBytes[:messageLen]
		copy(pack.MsgBytes, unframed)
```
Also if read_message("raw") is called on a new empty message it should not return MAX_MESSAGE_SIZE of NUL bytes (which is the reason the change is included here)

rafrombrc: One [place](https://github.com/mozilla-services/heka/blob/dev/pipeline/pipeline_runner.go#L215) where it happens is my fault, it just landed. I'm pretty sure I tried to make this change earlier and hit a side-effect somewhere else, too, but I just searched through the code and couldn't find anything.

I'm fine w/ trying this out, but the code I linked to above should be changed so the resize happens before the copy, and the Zero method should be changed to match the new pack behavior.

trink:
Have to make half that last change in 0.9 and the other in dev.  Closing and deleting this branch

Issue #1409, opened 2015-03-11
Matcher not clearing the regexp


trink:
'Logger =~ /./ && Type =~ /^foo/'

The second regexp is a special case and uses HasPrefix instead of an actual regexp.  The match expression builder wasn't clearing the previous regexp '.' when processing this special case and the matcher was using it.

PR #1410, opened 2015-03-12
Issue #1409 manually reset the message matcher parser struct with each c...

trink:
...all


trink:
should be against 0.9

Issue #1412, opened 2015-03-12
Retry config not being honoured on ProcessInputs where ticker_interval > 0


hynd:
Following on from #1395, if a ProcessInput is configured with a ticker_interval > 0, ```max_retries``` isn't honoured (either if the Process starts successfully and subsequently dies, or fails to start completely).

This minimal config is enough to illustrate on 0.9:
```
[ProcessInput]
ticker_interval = 10
[ProcessInput.command.0]
#bin = "/bin/true"
bin = "/bin/false"
#bin = "/bin/irmissing"
[ProcessInput.retries]
max_retries = 5
```

A ProcessInput with a non-zero exiting bin, or a missing bin gets respawned indefinitely.

PR #1413, opened 2015-03-12
Honour retries regardless of ticker_interval

hynd:
I've had a crack at fixing #1412, hopefully without breaking anything else unintentionally....

It seemed like ProcessInput's ```Run()``` needed to return any kind of error from ```runOnce()``` to the plugin runner, regardless of ticker_interval.  I added a var to push into any errors that were returned, which can then be punted back up to trigger the retry behaviour.

I close the stopChan in ```runOnce()```.  I also return early in some cases.  It looked like that's what would be wanted (and reduces a little duplicate logging).  Does that seem sane?

I tried it out with good (zero-exiting), bad (non-zero exiting) and missing ProcessInput ```bin=``` commands, with ticker_interval set to both 0 and >0, and max_retries set to -1 and >=0 and it appears to do the right thing in each case.

### DIFF
--- a/plugins/process/process_input.go
+++ b/plugins/process/process_input.go
@@ -17,7 +17,6 @@
 package process
 
 import (
-	"errors"
 	"fmt"
 	"github.com/mozilla-services/heka/message"
 	. "github.com/mozilla-services/heka/pipeline"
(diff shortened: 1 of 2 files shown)


hynd:
Per #1418, this changes the behaviour of ProcessDirectoryInput retries.

rafrombrc:
This looks reasonable, but it needs to have an entry in the CHANGES.txt changelog. Also, I'd love to see an update to the ProcessInput docs to clarify this behavior, explicitly stating that any time a process exits w/ an error it will cause the input to exit, triggering a shutdown or the retry behavior.

hynd:
Have another PR on the way that bundles this up with a couple of other tweaks.

Issue #1393, opened 2015-03-06
UdpOutput needs to check outgoing data size


rafrombrc:
Used to be MAX_MESSAGE_SIZE was hard-coded to 64KiB, which is typically the maximum size of a UDP packet, but now it's configurable. The UdpOutput should check to make sure the outgoing payload will actually fit into a UCP packet and record an error if the data is being truncated.

PR #1416, opened 2015-03-12
UdpOutput drops message if its size exceeds allowed UDP data size (#1393)

4r9h:
...).


4r9h:
@rafrombrc new PR for release 0.9.1 :)

Issue #1387, opened 2015-03-03
Add InfluxDB v0.9.0+ API support and features


acesaro:
When upgrading to the upcoming 0.9.0 release of InfluxDB, there are some major changes occurring in the API to write and read data to/from the database via HTTP.  There is also strong support and encouragement for the use of tags for improved query and aggregation capabilities.  The new API call structure for writing data to InfluxDB 0.9.0+ is documented here: http://influxdb.com/docs/v0.9/concepts/reading_and_writing_data.html#writing-data-using-the-http-api

This issue is to track progress on the upgrades to the InfluxDB encoders (Schema InfluxDB Encoder and StatMetric InfluxDB Encoder) and potentially adding a decoder for supporting the periodic querying of data from InfluxDB into Heka messages for various purposes.

Here are the proposed updates up for discussion.  I have already started building them using this as a plan, but I would appreciate any input on any better ways to implement.

* Add a new optional boolean config option called `post_v09_api` that switches the logic throughout the script that generates the message structure to support the InfluxDB 0.9.0+ API.  This will default to false to maintain the existing functionality by default.
* Add a new optional config option `database` that defines the database that the output of the encoders is destined to within InfluxDB.  This was formerly embedded in the `address` config option for the encoders but needs to be split out as it is part of the API call now.  As a result, the example values for `address` will be updated in documentation for InfluxDB 0.9.0 endpoints.
* Add a new optional config option `retention_policy` which will map to the retentionPolicy attribute in the InfluxDB 0.9.0 API call.
* Add a new optional config option `map_fields_as_tags` that automatically maps fields (base and dynamic) in the Heka message that make sense to be used as tags (e.g. non-numerical values).
* Add a new optional config option `tag_fields` that allows the explicit definition of fields to be included as tags, overriding the `map_fields_as_tags` option when it is defined as true.
* Potentially add a new config option `timestamp_precision` that formats the timestamp appropriately in the write API message.  This would have values that match what InfluxDB accepts (n, u, ms, s, m, and h) and does the math on the timestamp value accordingly.
* Potentially add a new decoder InfluxDB Query Decoder which takes the JSON in the response to an InfluxDB 0.9.0+ query request and injects a Heka message into the pipeline for desired routing.  The potential use case here is taking the result of an HTTP input which runs an aggregate query periodically and then routes it to filters to potentially use anomaly detection or routing to another monitoring system.

[...]

PR #1420, opened 2015-03-15
Add Schema InfluxDB Write Encoder

acesaro:
This is the first commit for functionality outlined in https://github.com/mozilla-services/heka/issues/1387, which updates compatibility with the InfluxDB 0.9.0+ APIs that have changed quite a bit in that release.


sandbox/lua/encoders/schema_influx_write.lua
```diff
@@ -0,0 +1,319 @@
...
+    "Pid", "Logger", "Severity", or "EnvVersion" will be extracted from the
+    the base message schema, any other values will be assumed to refer to a
+    dynamic message field. Only the first value of the first instance of a
+    dynamic message field can be used for name name interpolation. If the
+    dynamic field doesn't exist, the uninterpolated value will be left in the
+    name. Note that it is not possible to interpolate either the
+    "Timestamp" or the "Uuid" message fields into the name, those
+    values will be interpreted as referring to dynamic message fields.
+
+- database (string, required, default "mydb")
```
rafrombrc: A config setting being required should mean that no default is provided, but that the plugin will error out at startup time if the value isn't specified.

sandbox/lua/encoders/schema_influx_write.lua
```diff
@@ -0,0 +1,319 @@
...
+    The InfluxDB database to store the metrics in.
+
+- retention_policy (string, optional, default "")
+    The InfluxDB retentionPoilcy value for the database in which the
+    metrics are being sent to.  Only relevant when post_v09_api is true.
```
rafrombrc: Dangling reference to `post_v09_api` which is no longer relevant.

rafrombrc: Also, is it worth specifying what an empty value actually means, with regard to InfluxDB retention policy?

sandbox/lua/encoders/schema_influx_write.lua
```diff
@@ -0,0 +1,319 @@
...
+    dynamic field doesn't exist, the uninterpolated value will be left in the
+    name. Note that it is not possible to interpolate either the
+    "Timestamp" or the "Uuid" message fields into the name, those
+    values will be interpreted as referring to dynamic message fields.
+
+- database (string, required, default "mydb")
+    The InfluxDB database to store the metrics in.
+
+- retention_policy (string, optional, default "")
+    The InfluxDB retentionPoilcy value for the database in which the
```
rafrombrc: typo: retentionPoilcy -> retentionPolicy

sandbox/lua/encoders/schema_influx_write.lua
```diff
@@ -0,0 +1,319 @@
...
+    be assumed to refer to the corresponding field from the base message
+    schema. Any other values will be assumed to refer to a dynamic message
+    field. The magic value "**all_base**" can be used to exclude base fields
+    from being mapped to the event altogether (useful if you don't want to
+    use tags and embed them in the name_prefix instead).
+
+- tag_fields (string, optional, default "**all_base**")
+    Take fields defined and adds them as tags of the point(s) sent to
+    InfluxDB for the message.  The magic values "**all**" and "**all_base**"
+    are used to map all fields (including base) to tags and only base fields
```
rafrombrc: Is this actually "all fields" or should it be "all non-numeric fields"?

sandbox/lua/encoders/schema_influx_write.lua
```diff
@@ -0,0 +1,319 @@
...
+    to tags, respectively.  If those magic values aren't used, then only those
+    fields defined will map to tags of the points sent to InfluxDB.
+
```
rafrombrc: It's not quite clear from the docs so far whether `skip_fields` and `tag_fields` impact each other. If a field is added to `skip_fields` and `tag_fields`, will it show up in the tags?

sandbox/lua/encoders/schema_influx_write.lua
```diff
@@ -0,0 +1,319 @@
...
+- timestamp_precision (string, optional, default "n")
+    Specify the timestamp precision that you want the event sent with.  The
+    default is to use nanoseconds by dividing the Heka message timestamp
+    by 1000000, but this math can be altered by specifying one of the precision
+    values supported by the InfluxDB write API (n, u, ms, s, m, h).  The default
+    of nanoseconds is used to match the precision of Timestamp in Heka messages.
```
rafrombrc: This says the default is nanoseconds, but the code below seems to be using a default of milliseconds.

sandbox/lua/encoders/schema_influx_write.lua
```diff
@@ -0,0 +1,319 @@
...
+    Take fields defined and adds them as tags of the point(s) sent to
+    InfluxDB for the message.  The magic values "**all**" and "**all_base**"
+    are used to map all fields (including base) to tags and only base fields
+    to tags, respectively.  If those magic values aren't used, then only those
+    fields defined will map to tags of the points sent to InfluxDB.
+
+- timestamp_precision (string, optional, default "n")
+    Specify the timestamp precision that you want the event sent with.  The
+    default is to use nanoseconds by dividing the Heka message timestamp
+    by 1000000, but this math can be altered by specifying one of the precision
```
rafrombrc: This would be more readable with separators (i.e. "1,000,000") or, even better, as 1e6.

sandbox/lua/encoders/schema_influx_write.lua
```diff
@@ -0,0 +1,319 @@
...
+
+local name_prefix_orig  = read_config("name_prefix") or "name"
+local name_prefix = name_prefix_orig
+local use_subs
+if string.find(name_prefix, "%%{[%w%p]-}") then
+    use_subs = true
+end
+
+-- Default this option to "mydb"
+local influxdb_db = read_config("database") or "mydb"
```
rafrombrc: Here's where you enforce the database value:

```lua
local influxdb_db = read_config("database") or error("`database` setting required")

File: sandbox/lua/encoders/schema_influx_write.lua [hunk 9]
```diff
@@ -0,0 +1,319 @@
...
+        timestamp_divisor = 1e3
+    elseif timestamp_precision == "ms" then
+        timestamp_divisor = 1e6
+    elseif timestamp_precision == "s" then
+        timestamp_divisor = 1e9
+    elseif timestamp_precision == "m" then
+        timestamp_divisor = 1e9 * 60
+    elseif timestamp_precision == "h" then
+        timestamp_divisor = 1e9 * 60 * 60
+    end
```
rafrombrc: `timestamp_divisor` value doesn't change for the lifetime of the plugin, should be calculated outside of `process_message` rather than repeated for each message.

sandbox/lua/encoders/schema_influx_write.lua
```diff
@@ -0,0 +1,319 @@
...
+
+    -- Create a counter for the table indexes for columns and values
+    -- then populate those tables with the used base fields previously
+    -- determined
+    local place = 1
+    for _, field in ipairs(used_base_fields) do
+        columns[place] = field
+        values[place] = read_message(field)
+        place = place + 1
+    end
```
rafrombrc: You don't have to track `place` manually, you can just use `for place, field in ipairs(used_base_fields) do`.

sandbox/lua/encoders/schema_influx_write.lua
```diff
@@ -0,0 +1,319 @@
...
+    if base_fields_map[key] then
+        return read_message(key)
+    else
+        local val = read_message("Fields["..key.."]")
+        if val then
+            return val
+        end
+        return "%{"..key.."}"
+    end
+end
```
rafrombrc: Oof. I thought I had extracted the message_interpolation stuff out into its own separate module so the various sandbox plugins wouldn't have to reinvent this wheel again and again. Turns out I did do this work, but I did it on a branch that hasn't landed, for reasons completely unrelated to the message interpolation stuff.

The commit in question, which includes changes to existing plugins and documentation, can be found at 22501291d264572ba8348f0c75f881787b5b755d. It's a big ask, but might you be willing/able to extract the bulk of that commit and add it as a commit to your branch, and then change your code to use the shared implementation?

If that's too much for you (and I understand if it is), then I can pull that commit out and get it merged back to dev, and once that's landed you can rebase against dev and then update your code to use the shared implementation.

sandbox/lua/encoders/schema_influx_write.lua
```diff
@@ -0,0 +1,319 @@
...
+    else
+        name_prefix = name_prefix_orig
+    end
+
+    -- Initialize and populate the table of tags to include in the
+    -- InfluxDB write API message
+    -- Convert value to a string as this is required by the API
+    local message_tags = {}
+    for _, field in ipairs(base_fields_list) do
+        if (used_tag_fields["**all_base**"] or used_tag_fields["**all_base**"])
```
rafrombrc: You're or'ing the same value against itself here.

rafrombrc: Looking at this further, it occurs to me that you know whether the `**all_base**` and `**all**` keys exist at config loading time. We can convert these key lookups to boolean values and save a few cycles, rather than having to perform the key lookup multiple times inside this loop for every message.

sandbox/lua/encoders/schema_influx_write.lua
```diff
@@ -0,0 +1,319 @@
...
+            -- Increment the index to the table so the next iteration will
+            -- append another entry to the array. This allows us to send a
+            -- bunch of metrics to InfluxDB with a single HTTP request
+            points_index = points_index + 1
+        end
+
+        -- Include the dynamic fields as tags if they are defined in
+        -- configuration or the magic value "**all**" is defined.
+        -- Convert value to a string as this is required by the API
+        if used_tag_fields[name] or used_tag_fields["**all**"] then
```
rafrombrc: Here's another place where a boolean could be set at config loading time, and checked before the individual field lookup to avoid unnecessary hash checks.

sandbox/lua/encoders/schema_influx_write.lua
acesaro: I see that you just recently merged this into dev, so thanks!  I had this completed last night (along with good progress on the rest of the notes in your review), so I'll remove the changes outside of this file that I had pending to commit.  It works great, by the way. :)

rafrombrc: It was actually @RobCherry who did the work of teasing that stuff out... I assumed he was working with you, but I guess not? Anyway, thanks to both of you for helping out! :)

acesaro:
I've submitted a new commit that I believe clears up all the comments that you made.

One item that I would like some help with, as I could not figure it out quickly, is support for large integers to not be converted to scientific notation.  Currently, the "u" and "n" `timestamp_precision` values do not work properly when sent to InfluxDB as they are converted to scientific notation when the JSON encoding is done.  Does anyone have any suggestions on how to do this in Lua?  From my research it might take the inclusion of a non-packaged module to achieve.  If not, we can resort to excluding those values from the support of the plugin.  Thanks!

sandbox/lua/encoders/schema_influx_write.lua
```diff
@@ -0,0 +1,330 @@
...
+Tags, timestamp/precision, database, and retention policy are defined once
+for all points submitted in the API call.  All dynamic fields in the Heka
+message are converted to separate series in the `points` array that is
+submitted to InfluxDB.
+
+Config:
+
+- name_prefix (string, optional, default "name")
+    String to use as the `name` key's prefix value in the generated JSON.
+    Supports interpolation of field values from the processed message, using
```
rafrombrc: You can just refer to the message field interpolation docs here, like we do in the [cbuf_librato docs](https://github.com/mozilla-services/heka/blob/dev/sandbox/lua/encoders/cbuf_librato.lua#L24).

sandbox/lua/encoders/schema_influx_write.lua
```diff
@@ -0,0 +1,330 @@
...
+            -- Increment the index to the table so the next iteration will
+            -- append another entry to the array. This allows us to send a
+            -- bunch of metrics to InfluxDB with a single HTTP request
+            points_index = points_index + 1
+        end
+
+        -- Include the dynamic fields as tags if they are defined in
+        -- configuration or the magic value "**all**" is defined.
+        -- Convert value to a string as this is required by the API
+        if (used_tag_fields[name] or tag_fields_all)
```
rafrombrc: It's a small win, but it'd be better to put the `tag_fields_all` check first to skip the hash key lookup when we know we're using all fields.

rafrombrc:
This looks great, very close to being ready to merge, thanks for your dedication (and patience with my code review).
As for the timestamp precision issue... ugh. I see that cjson has an `encode_number_precision` setting, but it's max value is 14, and that's not enough. I think for now we might have to omit those precision options. We probably also want to file a ticket InfluxDB re: accepting scientific notation values, since that's a valid part of the [JSON spec](http://www.json.org/).

acesaro:
Updates have been made per the comments in your code review and above.  Thanks!

acesaro:
Crap, not sure what I did here but I somehow included other commits that I was just trying to fetch and merge from upstream into my branch.

acesaro:
Ok, it looks like I got it fixed.  Let me know if you see any more issues.  Thanks!

Issue #1418, opened 2015-03-13
ProcessDirectoryInput retry failure.


hynd:
Another retry-related one, sorry :(

Any retry config for a ProcessDirectoryInput managed ProcessInput is ignored.

ProcessInputs with a ticker_interval >0 (ie; in a directory >0) that exit with a non-zero status are respawned, but not because of the retry config, but because the fall off the end of process_input's ```runOnce()``` and the RunCmd tickChan starts it again.
That might be as intended? If it is, PR #1413 will change that behaviour since a failure will cause the ProcessInput's ```Run()``` to return, which ends up doing the same as below... (ie; never respawning).

If a long-running ProcessInput (ie; in a 0 directory) exits, it never respawns, regardless of any retry config (changing the config file does trigger a respawn).

ProcessInput's registered by PDI are set as ["transient"](https://github.com/mozilla-services/heka/blob/adcc1d132bece8e5e42bc1ad78b6614d3e3ea286/plugins/process/process_directory_input.go#L221) which looks like it causes them to [break out](https://github.com/mozilla-services/heka/blob/adcc1d132bece8e5e42bc1ad78b6614d3e3ea286/pipeline/plugin_runners.go#L306) of the ```ir.Starter()``` for loop here before it ever gets to the ```rh.Wait()```?

If that happens, AFAICT, ProcessDirectoryInput doesn't handle any retries and won't know to remove it from the pdi.inputs map and merrily [continues](https://github.com/mozilla-services/heka/blob/adcc1d132bece8e5e42bc1ad78b6614d3e3ea286/plugins/process/process_directory_input.go#L136) without spawning it again.

If i fudge plugin_runners so transients do successfully hit the ```rh.Wait()```, any unclean ProcessInput exits do trigger the retry behaviour, but it seems like any retry config gets clobbered by SetCommonTypedConfig on [#L213](https://github.com/mozilla-services/heka/blob/adcc1d132bece8e5e42bc1ad78b6614d3e3ea286/plugins/process/process_directory_input.go#L213) running after the ```loadProcessFile``` on [#L200](https://github.com/mozilla-services/heka/blob/adcc1d132bece8e5e42bc1ad78b6614d3e3ea286/plugins/process/process_directory_input.go#L200), so retries defaults to -1 and it goes on forever.

I think i have some fixes which make it behave as i think i understand it should, but i may have misinterpreted the docs :)  Any clarifications or pointers would be appreciated!

rafrombrc:
[...]

PR #1428, opened 2015-03-23
Retry logic changes for Inputs

hynd:
* Honour retry config regardless of ticker_interval, or whether spawned from a
  ProcessDirectoryInput (for #1412 and #1418)
* Plugins exiting cleanly will not be subject to Retries
* Make InputRunners support can_exit (defaults to false, ProcessDirectoryInput
  spawned processes default to true).  Skipped if runner was removed
  deliberately (ie; on a ProcessDirectoryInput config change)
* ProcessInput spawned commands pass exit error back up to plugin_runner.
* Reset the stopChan, once mutex, exitError and cc in Run rather than Init (so
  ProcessDirectoryInput commands spawned after a config change have a fresh
  start)
* Doc + some test updates (could do with more improvement)

### DIFF
--- a/docs/source/developing/plugin.rst
+++ b/docs/source/developing/plugin.rst
@@ -147,19 +147,16 @@ generally route messages to specific outputs using the
 Restarting Plugins
 ==================
 
-In the event that your plugin fails to initialize properly at startup, hekad
-will exit. However, once hekad is running, if the plugin should fail (perhaps
-because a network connection dropped, a file became unavailable, etc) then the
-plugin will exit. If the plugin supports being restarted then Heka will
-attempt to reset, reinitialize, and restart the plugin. If this fails, Heka
-will try again up until the specified max_retries value. If the failure
-continues beyond the maximum number of retries, or if the plugin didn't
-support restarting in the first place, then Heka will either shut down or, if
-the plugin is a filter or an output with the ``can_exit`` setting set to true,
-the plugin will be removed from operation and Heka will continue to run.
+If your plugin supports being restarted and either fails to initialize properly
+at startup, or fails during Run with an error (perhaps because a network
+connection dropped, a file became unavailable, etc) then Heka will attempt to
+reinitialize and restart it up until the specified max_retries value.
 
-If the reinitialization and restarting is successful, then the retry count
-will be reset to zero and everything will continue to function as normal.
+If the failure continues beyond the maximum number of retries, or if the plugin
+didn't support restarting in the first place, then Heka will either shut down
+or, if the plugin is an input, filter or an output with the ``can_exit``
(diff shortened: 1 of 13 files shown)


docs/source/developing/plugin.rst
```diff
@@ -147,19 +147,16 @@ generally route messages to specific outputs using the
...
-continues beyond the maximum number of retries, or if the plugin didn't
-support restarting in the first place, then Heka will either shut down or, if
-the plugin is a filter or an output with the ``can_exit`` setting set to true,
-the plugin will be removed from operation and Heka will continue to run.
-
-If the reinitialization and restarting is successful, then the retry count
-will be reset to zero and everything will continue to function as normal.
+If your plugin supports being restarted and either fails to initialize properly
+at startup, or fails during Run with an error (perhaps because a network
+connection dropped, a file became unavailable, etc) then Jeka will attempt to
```
rafrombrc: Typo (Jeka -> Heka)

hynd:
Doh, fixed.  Tidied a couple of other comments i'd put in the code itself too.

pipeline/plugin_runners.go
```diff
@@ -279,59 +288,77 @@ func (ir *iRunner) Starter(h PluginHelper, wg *sync.WaitGroup) {
...
-		}
 
-		// Are we supposed to stop? Save ourselves some time by exiting now.
-		if globals.IsShuttingDown() {
-			break
-		}
+			// Otherwise we'll execute the Retry config
+			recon.CleanupForRestart()
+			if ir.maker == nil {
+				ir.maker = ir.pConfig.makers["Input"][ir.name]
```
rafrombrc: PipelineConfig.makers is protected by a RWLock, you should RLock and RUnlock around your usage. (Bootstrap code doesn't do so b/c there's only one running goroutine at that point.)

rafrombrc: And I realize you've copied this code from the foRunner loop... it has the same error. :P

pipeline/plugin_runners.go
```diff
@@ -279,59 +288,77 @@ func (ir *iRunner) Starter(h PluginHelper, wg *sync.WaitGroup) {
...
-		}
-		// We're not transient, we either try to restart or we shut down
-		// altogether.
-		recon, ok := ir.plugin.(Restarting)
-		if !ok {
-			ir.LogMessage("has stopped, shutting down.")
-			globals.ShutDown()
-			break
-		}
+		initLoop:
```
rafrombrc: And another bug that you've brought in from the foRunner code... we should be checking for globals.IsShuttingDown() in the initLoop, otherwise we might end up stuck here forever.

rafrombrc:
I'll go ahead and fix up these last issues and get this merged, no action needed.

Issue #1162, opened 2014-11-04
Documentation about how to use UNIX DGRAM (/dev/log)


gdm85:
I recently discovered commit dd3db14bea8d4a222cb13969fe5ae9861219ef75 and #790, however I am having some issues at using it.

For example, if I specify `net = unixdgram` I receive an error:

```
Initialization failed for 'UdpInput': ResolveUDPAddr failed: unknown network unixdgram
```

However documentation [here](https://hekad.readthedocs.org/en/latest/config/inputs/#udpinput) says:

```
New in version 0.5.
    net (string, optional, default: “udp”)
        Network value must be one of: “udp”, “udp4”, “udp6”, or “unixgram”.
```

I think an example would help, I did not contribute any because I am still tinkering with it.

Issue #790, opened 2014-04-14
Add DGRAM support for unix sockets to create /dev/log


mmerickel:
`/dev/log` is the socket used by syslog and subsequently the `logger` command. It uses a DGRAM-configured unix socket, rather than the currently supported streaming socket.. Can I ask why this isn't already supported by heka? As a logging library I would've thought that syslog support would be one of the first features added as it is used for auth/sudo/postfix and many other system services and the lack of the feature does make me question whether I'm thinking of heka as a proper replacement for rsyslog and other logging solutions.

PR #1440, opened 2015-03-27
Raw syslog datagrams multidecoder example

gdm85:
I am proposing to add an example to cover #1162 and the older #790.

This is particularly interesting in the case of containers, if one does not want to run syslog/rsyslog inside the container but rather directly read ``/dev/log`` via heka.

Although the best scenario is to let syslog/rsyslog poll ``/dev/log`` and write their own log (which is then feeded to/read by heka), I prefer this approach and I provide here a working example (tested with 0.9.0) that should also allow people to easily extend/build on.

I am new to heka, so please tell me about possible improvements. Feedback for PR changes welcome as well :)

**NOTE:** might require some refactoring of the name of sibling example.toml, if this is accepted at all

1. Is there already a decoder that does this?
2. If not, would this better be converted to a LUA decoder?

I didn't write (2) as I wanted a simple message proxying feature, but for official inclusion it might be a valid option instead of a multidecoder example (also to properly parse PRI).

### DIFF
--- a/examples/raw_syslog_unixgram.toml
+++ b/examples/raw_syslog_unixgram.toml
@@ -1,0 +1,44 @@
+[syslog_udp]
+type = "UdpInput"
+net = "unixgram"
+address = "/dev/log"
+decoder = "RawSyslogDecoder"
+
+[RawSyslogDecoder]
+type = "MultiDecoder"
+subs = ['SyslogDecoderWProcess', 'SyslogDecoderSimple']
+cascade_strategy = "first-wins"
+## some extra verbosity here when matching the simple decoder (as the first fails with error)
+log_sub_errors = true
+
+[SyslogDecoderWProcess]
+type = "PayloadRegexDecoder"
+## <86>Mar 27 11:18:23 su[975952]: pam_unix(su:session): session opened for user root by (uid=0)
+match_regex = '^<(?P<Pri>\d+)>(?P<Timestamp>[A-Za-z]{3} \d{2} \d+:\d+:\d+) (?P<Payload>((?P<ProcessName>[^\[:]+)\[(?P<Pid>\d+)\]: )?.+)$'
+timestamp_layout = "Jan 02 15:04:05"
+# timestamp_location = "Europe/Amsterdam" ## specify here the timezone, as timestamps are provided in local time
+
+[SyslogDecoderWProcess.message_fields]
+Pid = "%Pid%"
+Payload = "%Payload%"
+ProcessName = "%ProcessName%"
+## see http://en.wikipedia.org/wiki/Syslog#Severity_levels for properly decoding the PRI field
+Pri = "%Pri%"
+
+[SyslogDecoderSimple]
+type = "PayloadRegexDecoder"
+## <86>Mar 27 11:18:23 root: hello
+match_regex = '(^|\n)<(?P<Pri>\d+)>(?P<Timestamp>[A-Za-z]{3} \d{2} \d+:\d+:\d+) (?P<Payload>[^\n]+)'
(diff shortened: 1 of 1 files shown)


gdm85:
@rafrombrc the provided example should cover http://tools.ietf.org/html/rfc5424#section-6

However, if you say that there is already a decoder covering RFC5424 (that I implemented in a poor way, I have to admit), then I'll withdraw the PR :)

steverweber:
+1

just opened a issue then found this PR.
https://github.com/mozilla-services/heka/issues/1573

Just missing the output decoder.

steverweber:
logstash has a syslog input... thats kinda what I'm looking to support from heka,
https://www.elastic.co/guide/en/logstash/current/plugins-inputs-syslog.html

gdm85:
@steverweber the idea here was to provide a basic working example, then others can improve on top of it. Although didn't receive any feedback yet...

cristi1979:
I just want to say that our experience with heka parsing of /dev/log was not very good.

We built a lua script instead of using regex, but we quickly learned that messages can take _ANY_ form, there is no standard forced on what reached /dev/log. We concluded that we don't want to reimplement all the logic that makes rsyslog/syslog-ng good log parsers in lua.

So we have a local rsyslog server that takes care of /dev/log and forwards all messages in a rsyslog standard format to a local heka server, where we do the processing. Heka also parses various log files that are not sent via /dev/log.

gdm85:
@cristi1979 yes, I agree with you - having heka take care of ``/dev/log`` would probably be a too big leap

steverweber:
correct me if im wrong having an UdpOutput formatter that does one of the simple syslog styles should be straight forward?

I'm looking to have heka parse log files then send them to this logstash plugin using the syslog format. (its how the IT staff seem to be collecting logs)
https://www.elastic.co/guide/en/logstash/current/plugins-inputs-syslog.html

@cristi1979 perhaps you have an example Encoder.

Thanks.

gdm85:
@steverweber this PR is about an example in official release explaining how to have heka receive syslog on a listening UNIX socket e.g. ``/dev/log`` (although I agree with the reserves mentioned by @cristi1979 , that it will never be as good as a proper dedicated daemon), so it's about harvesting logs without having a syslog daemon running. Seems like what you're looking for is related to output, not input.

cristi1979:
I've put here what we did: https://gist.github.com/cristi1979/b0a8b6e6d312096f9515

The bulk logic was inspired by the rsyslog plugin.

trink:
@cristi1979 You can get rid of over half that code if you use the grammar generator.

cristi1979:
Hi @trink, can you help me with some pointers on that? I will start writing a parser for glassfish logs and maybe I can do better with that one.

gdm85:
It's almost 1 year this PR has been open, is it going to be merged or..?

rafrombrc:
Sorry, clearly I've been neglectful. I'm afraid this isn't really the approach we want to encourage. The PayloadRegexDecoder hasn't been removed from the tree because sometimes it fills a gap for some folks, but we generally don't recommend its use; a proper LPEG grammar in a SandboxDecoder is nearly always a better choice.

In (slightly) better news, PR #1760 contains an implementation of a generic syslog decoder which is meant to be able to attach to a syslog stream and parse all of the log data coming through. That PR has also been languishing, however, I'll bump it up higher in my queue and will try to get it merged by the end of this month.

Sorry again for not responding.

gdm85:
@rafrombrc ah, no problem! I was keeping around the fork only for this PR, but glad to hear there is a better way to achieve this. I see it as an important feature when comparing to alternatives like fluentd etc

PR #1443, opened 2015-03-30
a gziped file seeking in log streamer may cause an OOM exception

chzyer:
For the old codes, in some case, like `seekPos` is large enough, may cause an OOM exception.
And then, `ioutil.Discard` will grab a `[]byte` from `blackHolePool`, it may relieve gc's pressure,

### DIFF
--- a/logstreamer/reader.go
+++ b/logstreamer/reader.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 
 	"github.com/mozilla-services/heka/ringbuf"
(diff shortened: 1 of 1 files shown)


rafrombrc:
This is great, thanks! I'd love to have this fix land in the 0.9.2 release, would you be willing to close this PR and open a different one against the `versions/0.9` branch, adding an entry to the "Bug Handling" section for 0.9.2 in the CHANGES.txt changelog?

If you can't get to this in the next 24 hours or so that's fine, just please let me know so I can take care of it.

chzyer:
move to: https://github.com/mozilla-services/heka/pull/1447

PR #1447, opened 2015-03-30
Fix a gziped file seeking in log streamer may cause an OOM exception

chzyer:
For the old codes, in some case, like `seekPos` is large enough, may cause an OOM exception.
And then, `ioutil.Discard` will grab a `[]byte` from `blackHolePool`, it may relieve gc's pressure,

Landing in the 0.9.2 release.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,12 @@
 0.9.2 (2015-??-??)
 ==================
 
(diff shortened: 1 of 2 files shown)


chzyer:
It's landing in the 0.9.2 release now! @rafrombrc

PR #1449, opened 2015-03-31
Handle chunked error responses in http_output

hynd:
Some servers choose to send chunked responses which have no Content-Length header (ie; one implemented in Go will default to this unless you set a Content-Length header yourself).


hynd:
Will reopen against versions/0.9

PR #1457, opened 2015-03-31
ElasticSearch bulk api expects newline in the end of the payload

aaronfeng:
Below is the tcpflow output without the newline:

```
POST /_bulk HTTP/1.1
Host: 172.17.0.36:9200
User-Agent: Go 1.1 package http
Content-Length: 97
Accept: application/json
Accept-Encoding: gzip
 
{"index":{"_index":"2015.03.31","_type":"7d98073d388e"}}
{"bar":"boo","log":"awesome"}
HTTP/1.1 400 Bad Request
Content-Type: application/json; charset=UTF-8
Content-Length: 99
 
{"error":"ActionRequestValidationException[Validation Failed: 1: no requests added;]","status":400}
```

### DIFF
--- a/sandbox/lua/encoders/es_payload.lua
+++ b/sandbox/lua/encoders/es_payload.lua
@@ -78,7 +78,7 @@ function process_message()
 
     -- bulk api expects newline in the end of the payload
     if not string.match(payload, "\n$") then
-	      payload = payload .. "\n"
+        add_to_payload("\n")
     end
 
     add_to_payload(idx_json, "\n", payload)
--- a/sandbox/lua/encoders/es_payload.lua
+++ b/sandbox/lua/encoders/es_payload.lua
@@ -72,8 +72,16 @@ function process_message()
     if ts_from_message then
         ns = read_message("Timestamp")
     end
+
     local idx_json = elasticsearch.bulkapi_index_json(index, type_name, id, ns)
-    add_to_payload(idx_json, "\n", read_message("Payload"))
+    local payload  = read_message("Payload")
+
+    -- bulk api expects newline in the end of the payload
+    if not string.match(payload, "\n$") then
+	      payload = payload .. "\n"
+    end
+
+    add_to_payload(idx_json, "\n", payload)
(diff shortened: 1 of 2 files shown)


sandbox/lua/encoders/es_payload.lua
```diff
@@ -72,8 +72,16 @@ function process_message()
...
     end
+
     local idx_json = elasticsearch.bulkapi_index_json(index, type_name, id, ns)
-    add_to_payload(idx_json, "\n", read_message("Payload"))
+    local payload  = read_message("Payload")
+
+    -- bulk api expects newline in the end of the payload
+    if not string.match(payload, "\n$") then
+	      payload = payload .. "\n"
+    end
```
rafrombrc: Since Lua strings are immutable, this does some unnecessary allocation and will likely cause unnecessary GC (see http://stackoverflow.com/questions/1405583/concatenation-of-strings-in-lua). You can avoid this by using `add_to_payload`:

```lua
local payload = read_message("Payload")
add_to_payload(idx_json, "\n", payload)
if not string.match(payload, "\n$") then
    add_to_payload("\n")
end
inject_payload()
```

aaronfeng:
@rafrombrc thanks.  fixed.

sandbox/lua/encoders/es_payload.lua
```diff
@@ -72,8 +72,16 @@ function process_message()
...
     local idx_json = elasticsearch.bulkapi_index_json(index, type_name, id, ns)
-    add_to_payload(idx_json, "\n", read_message("Payload"))
+    local payload  = read_message("Payload")
+
+    -- bulk api expects newline in the end of the payload
+    if not string.match(payload, "\n$") then
+        add_to_payload("\n")
+    end
+
+    add_to_payload(idx_json, "\n", payload)
```
rafrombrc: You actually put this in the wrong order, we want the final "\n" to be added to the payload *after* the first batch of data. I've fixed it up and will merge, however.

PR #1475, opened 2015-04-10
ticker_interval in Stat Accumulator Input be greater than 0

### LINKED ISSUE #1474: Should ticker_interval in Stat Accumulator Input be greater than 0 ?
I suppose ticker_interval in Stat Accumulator Input should be greater than 0. 

There is no check for the input value. 

Mistakenly setting this value to 0 would break the Accumulator, and possibly divided by 0 error. 

I will fix it and send a pull request if it's a valid issue.

imormo:
I suppose ticker_interval in Stat Accumulator Input should be greater than 0.
Mistakenly setting this value to 0 would break the Accumulator, and possibly divided by 0 error.
This pull requests resolves #1474.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -47,6 +47,9 @@ Bug Handling
 * ProcessInput fixed to no longer leak decoder goroutines when reconfigured
   via ProcessDirectoryInput-driven config changes (#1444).
 
+* Check configuration, ticker_interval in Stat Accumulator Input be greater 
+  than 0 (#1474).
+
 0.9.2 (2015-??-??)
 ==================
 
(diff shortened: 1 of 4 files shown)


rafrombrc:
This is a nice fix, thanks. Before I can merge it, however, we need to have an entry added to the 'bug handling' section for 0.10.0 in the CHANGES.txt changelog, and we need to fix the test that is now failing. Looks like the test is failing b/c the error you added is now getting raised somewhere, should be easy to fix by changing the offending test to have a non-zero ticker interval.

imormo:
I am modifying the testing cases. Will commit a fix after it passes all testing cases.

imormo:
Will reopen a new pull request.

PR #1478, opened 2015-04-12
Added handling of deliver_incomplete_final to the SplitBytes() in SpliterRunner. Fixes #1455

### LINKED ISSUE #1455: Packet-based inputs drop records w/ no trailing delimiter
If a packet-based input is used with a TokenSplitter and there's no delimiter at the end of the message then the last record in the message will be dropped. If there's no delimiter in the message at all, then an error will be returned since no record will be found. For instance, if an AMQPInput using a TokenSplitter with `,` as the delimiter receives a message containing `this,is,a,set,of,records`, then `records` will be dropped. If it contains `this is a single record` then an error will be raised. This is unexpected, ideally we'd at least default to considering the trailing content to be a record.

4r9h:
@rafrombrc, @trink  could you look at this patch?

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -47,6 +47,8 @@ Bug Handling
 * ProcessInput fixed to no longer leak decoder goroutines when reconfigured
   via ProcessDirectoryInput-driven config changes (#1444).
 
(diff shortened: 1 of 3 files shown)


4r9h:
Closing because of https://github.com/mozilla-services/heka/pull/1480

PR #1479, opened 2015-04-13
Never retry a buffered record for ElasticSearch over HTTP if we

rafrombrc:
know the HTTP request went through.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -19,6 +19,10 @@ Bug Handling
 
 * Handle chunked error responses in http_output.
 
+* ElasticSearchOutput buffering now only retries messages when it's clear that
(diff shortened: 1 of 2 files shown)


k2xl:
I see you check if response is > 304.

Which responses does ES return that are 500 when posting data?

I think i have seen ES return 502 for when elastic search is still starting up - and I imagine we would want to continue trying messages that see that response.

rafrombrc:
@k2xl My understanding is that when the HTTP request is successful but Lucene doesn't like the JSON to be indexed ElasticSearch will return a 500 response code. Do you know of any documentation that clearly specifies which response code ES will return, and in what conditions? If so, I'd be happy to adjust this to do the right thing as often as possible.

Issue #1482, opened 2015-04-15
LogstreamerInput option to start at stream tail


rafrombrc:
We've had a couple of requests for LogstreamerInput to be able to skip the backlog of a stream and start at the tail instead. I'm thinking we can add an `initial_tail` boolean option, such that if `initial_tail = true` *and* there is no cursor file for the stream, the backlog will be ignored and processing will start at the end. If the cursor file exists, it will still be honored.

rafrombrc:
Fixed in #1831.

PR #1486, opened 2015-04-16
initial tail

hoffoo:
this is a shot at #1482


logstreamer/filehandling.go
```diff
@@ -471,7 +475,11 @@ func (ls *LogstreamSet) ScanForLogstreams() (result []string, errors *MultipleEr
 			if err != nil {
 				errors.AddMessage(err.Error())
 				position.Reset()
+				initialTail = ls.initialTail
+			} else if position.SeekPosition == 0 {
```
hoffoo: This is pretty bad, the reason is though that LogstreamLocationFromFile will not return an error if journal is missing. Maybe there is a better way to do this?

logstreamer/filehandling.go
```diff
@@ -437,7 +439,8 @@ func (ls *LogstreamSet) GetLogstreamNames() []string {
 func (ls *LogstreamSet) ScanForLogstreams() (result []string, errors *MultipleError) {
 	var (
 		logstream *Logstream
-		ok        bool
+
+		ok bool
```
rafrombrc: Why add the empty line?

logstreamer/filehandling.go
rafrombrc: Yes, you'd want to check for empty `Filename` and `Hash` attributes as well. I'd add an `IsZero()` method to the LogstreamLocation struct and that checks all three values and use `else if position.IsZero()` instead.

rafrombrc: Actually, scratch that. I think it'd make more sense for the `LogstreamLocationFromFile` to accept the initialTail value as an argument. If initialTail == true and there's no journal file it should return the LogstreamLocation value corresponding to the end of the stream.

rafrombrc:
Oops, I see you made the change recommended by one of my notes, but which was actually overridden by a subsequent note that gives us what I think is a cleaner result. Sorry about that. :(

In any event, overall this looks pretty good. Before merging it needs the change that I spelled out in my second note (i.e. do all the work in LogstreamLocationFromFile instead of checking for a zero value after the return). Also, the LogstreamerInput documentation needs to be updated, and an entry needs to be added to the new features section of the CHANGES.txt changelog for 0.10.

Thanks!

hoffoo:
@rafrombrc okay i hope the wording is okay, not very good at writing documentation.

please let me know if there are other issues!

hoffoo:
I really don't know why that test is failing, they aren't locally for me. Ill figure it out and fix it

tianlin:
Hi,hoffoo.I think you should pass filename to InitialTailSetPos instead of JournalPath.

rafrombrc:
I'm seeing the exact same failure as Travis when I run the tests locally... are you still not seeing the failure, @hoffoo?

rafrombrc:
I've dug into this a bit and have found that the feature as implemented doesn't work, and the rabbit-hole is a bit too deep for me to go down right now. Closing this PR, leaving the issue open so that we can hopefully come back to it soon.

PR #1491, opened 2015-04-20
Add docker event stream input

phobologic:
Tried to follow the method set out in DockerLogInput as closely as I
could, though it ended up being a bit simpler to implement since I only
need to connect to the docker API.

Wasn't sure the best way to test this - it honestly doesn't have much
logic to it.


phobologic:
:) Of course I submit when it seems like github is having issues... 

`fatal: unable to access 'https://github.com/AdRoll/goamz/': Operation timed out after 300009 milliseconds with 0 out of 0 bytes received`

docs/source/config/inputs/docker_event.rst
```diff
@@ -0,0 +1,48 @@
+.. _config_docker_event_input:
+
+Docker Log Input
```
rafrombrc: s/Log/Event/

phobologic:
Thanks - updated.

PR #1494, opened 2015-04-22
Check for out of bound index on truncated lines.

cheribral:
pipePos and colonPos can end up being the same when
encountering a truncated line, but this isn't caught which causes hekad
to crash.
This adds a check for that condition which will simply skip the line
instead of crashing.


rafrombrc:
I've added a changelog entry and merged this back to the versions/0.9 branch so it can land in the 0.9.2 release. Thanks!

PR #1510, opened 2015-05-01
Allow disabling name_prefix in schema_influx_write

mattharden:


sandbox/lua/encoders/schema_influx_write.lua
```diff
@@ -33,6 +33,7 @@ Config:
     name. Note that it is not possible to interpolate either the
     "Timestamp" or the "Uuid" message fields into the name, those
     values will be interpreted as referring to dynamic message fields.
+    You can disable the name prefix by setting its value to "**none**".
```
rafrombrc: I think it makes more sense to change the default from "name" to "" and disable the name prefix if they leave it blank. Note that this would mean you should make a note in the "Breaking Changes" section of the CHANGES.txt changelog for the 0.10.0 release.

sandbox/lua/encoders/schema_influx_write.lua
```diff
@@ -129,12 +130,17 @@ require "string"
 local interp = require "msg_interpolate"
 
 local name_prefix_orig  = read_config("name_prefix") or "name"
-local name_prefix = name_prefix_orig
```
rafrombrc: By getting rid of this variable definition you've made the scope of the `name_prefix` defined below to be global. You'll want to at least keep a `local name_prefix` declaration here.

sandbox/lua/encoders/schema_influx_write.lua
```diff
@@ -129,12 +130,17 @@ require "string"
...
+if string.find(name_prefix_orig, "%%{[%w%p]-}") then
     use_subs = true
 end
 
+if name_prefix_orig == "**none**" then
+    name_prefix_orig = ""
+else
+    name_prefix_orig = name_prefix_orig.."."
+end
+
```
rafrombrc: The two changes above can be replaced by:

```lua
local use_subs
if name_prefix_orig != "" then
    if string.find(name_prefix_orig, "%%{[%w%p]-}") then
        use_subs = true
    end
    name_prefix_orig = name_prefix_orig.."."
end
```

Then the rest of the code below should work as desired.

mattharden:
@rafrombrc: Do you want me to incorporate your suggestions and submit another pull request?

rafrombrc:
@mattharden No need, actually. This encoder is going away entirely, to be replaced by one that implements the newer InfluxDB "Line" protocol, as supported in #1595. I'll close this one now. Thanks for checking!

PR #1513, opened 2015-05-05
Improving deb packaging with init scripts and user creation

sathieu:


sathieu:
Replaces #1085

examples/conf/hekad.toml
```diff
@@ -0,0 +1,9 @@
+[hekad]
+maxprocs = 2
+
+[TcpInput]
+address = "127.0.0.1:5565"
+parser_type = "message.proto"
+decoder = "ProtobufDecoder"
```
rafrombrc: `parser_type` is no longer a supported config option, it's been replaced by splitter plugins. If you just remove the `parser_type` and `decoder` settings entirely, Heka will choose the right defaults.

rafrombrc:
Overall this looks great, and tested well, thanks for cleaning this up! The only thing, other than the config tweak I already commented on, is that we want to add something to the CHANGES.txt changelog about this. I think there should be an entry in the 'backwards incompatibility' section for the 0.10.0 release explaining that the debian packaging setup is now different, ideally explaining which prerequisite debian packages need to be installed before the new deb generation code will work.

Thanks!

sathieu:
@rafrombrc : I have removed the two lines from the config, and added 2 entries in CHANGES.txt.

sathieu:
I've also removed the DEBHELPER check from CMakeLists.txt, as it is in cmake/CMakeDebHelper.cmake with an explicit message (instead of silently dropping the deb target).

sathieu:
One more: I removed  debian/heka.default.in, as it was not used.

I think I'm done with this PR.

Issue #1543, opened 2015-05-19
Feature Request: Require Go 1.4 or greater


michaelgibson:
Resolves build dep in: https://github.com/mozilla-services/heka/pull/1540

PR #1544, opened 2015-05-19
Update docs and build files to require Go 1.4.

michaelgibson:
https://github.com/mozilla-services/heka/issues/1543


rafrombrc:
You'll need to bump the go revision in the .travis.yml so TravisCI will use 1.4. when it tries to build. :)

Issue #1545, opened 2015-05-20
Allow to overwrite the ContainerName for Docker log input


fernandezvara:
Heka DockerLogInput uses ID and Name from the container itself, but sometimes the container name is not the best way to match the logs with services.

For example, using Mesos and Marathon to spin up Docker instances, Marathon_Task_ID are the significant 'name' that will allow to match all logs for an application, since Marathon can't control the container name that will be created.

On creation, Marathon adds an Environment Variable to the container with the MESOS_TASK_ID, so it must overwrite the ContainerName with that one to be easily consumed.

PR #1546, opened 2015-05-20
Allow to overwrite the ContainerName using a Container Environment Variable. Issue 1545

fernandezvara:
This pull request allow to set in the DockerLogInput a configuration to set the ContainerName using an arbitrary environment variable defined inside the container. Refers to issue https://github.com/mozilla-services/heka/issues/1545

If the environment variable defined on the configuration is not set, it uses the ContainerName instead.

Added on the configuration a new entry name_from_env_var that allows to select the desired variable to use.

Sample configuration: 

(This sample is useful to get the mesos task id instead of the container name, since using Mesos/Marathon the container name does not match the marathon id)

```toml
[hekad]
maxprocs = 2

[Dashboard]
type = "DashboardOutput"
address = ":4352"
ticker_interval = 15

[DockerLogInput]
name_from_env_var = "MESOS_TASK_ID"

[influxdb]
type = "SandboxEncoder"
filename = "lua_encoders/schema_influx.lua"

    [influxdb.config]
    series = "docker"
    skip_fields = "Pid EnvVersion Severity Type"

[InfluxOutput]
message_matcher = "Type == 'DockerLog'"
encoder = "influxdb"
type = "HttpOutput"
address = "http://10.10.10.10:31002/db/logs/series"
username = "root"
password = "root"
```

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -57,6 +57,8 @@ Features
 
 * Added `graphite` module with helpers allowing to generate graphite metrics for counters and timeseries (#1461).
 
+* Allow to overwrite ContainerName using a Container environment variable for the DockerLogInput. (#1545)
+
 Bug Handling
 ------------
 
--- a/docs/source/config/inputs/docker_log.rst
+++ b/docs/source/config/inputs/docker_log.rst
@@ -47,7 +47,7 @@ Config:
 .. versionadded:: 0.10
 
 - name_from_env_var(string):
-		Overwrite the ContainerName with this environmental variable on the Container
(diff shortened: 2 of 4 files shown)


rafrombrc:
This looks great, thanks! Before I can merge it there needs to be an entry added to the 'features' section for version 0.10.0 of the CHANGES.txt changelog, and the docker_log.rst documentation file needs to be updated to include the new setting.

fernandezvara:
I hope it meets the documentation style. Thanks for accepting the request.

Issue #1511, opened 2015-05-02
[ProcessInput] capture return code


rishid:
For ProcessInput and ProcessDirectoryInput input plugins, is it possible to capture the return code of the process that gets executed?

I could then start using Nagios-like scripts and have heka execute my check_* scripts and then alert if rc == 2

rafrombrc:
I'd love to support this. Unfortunately, it's quite a bit trickier than you might expect. First, ProcessInput supports piping multiple commands together, so you'd have to make sure you handle all of the cases where an error happens somewhere along the chain. Second, and more importantly, process return codes don't work similarly across all platforms.

Even so, it would be possible to check whether the user is on a Unix-like OS, and, if so, capture the exit code from the last executed process in the chain, storing it in a message field for any stdout and/or stderr messages that are generated. It's a bit of work, though, and it's not at the top of our priority queue right now, so it will either have to wait a bit or be done by an outside contributor.

imormo:
I am working on this feature for my current project. 
I will contribute my implementation back.

PR #1547, opened 2015-05-22
ProcessInput resturn exit status

imormo:
Pull request for #1511.  I am sending this pull request for early feedback on the design. 
* ProcessInput no longer exits for subcommand errors, it stores the return code and subcommand error in two fields. 
* ProcessInput adds two fields 
** ExitStatus: return code of the last sub command if there is any. I assume that in most cases users only care about the return code of the last command in the pipe. 
** SubcmdErrors: Collection of error from each sub command, in the format of "Subcommand[<command number>] returned an error: <error> "

If this design is approved I will update document accordingly.


rafrombrc:
Overall this approach seems sound. There's some commented code left in there that needs to be removed. The only real concern I have is related to the syscall.WaitStatus stuff in the initDelivery function. Will that work on all platforms (i.e. Linux, OSX, Windows)? If it doesn't work on Windows, will it fail over gracefully?

imormo:
syscall.WaitStatus works for all platforms, to my best knowledge. 
There is the description, sys() is a platform dependent exit information.
https://golang.org/pkg/os/#ProcessState.Sys

Exec implementation for windows also indicates it implements all interfaces. 
https://golang.org/src/os/exec_windows.go

PR #1555, opened 2015-05-28
[issue #1554] fix `. build.sh` fails in 0.9.2 tarball release

### LINKED ISSUE #1554: `. build.sh` fails in 0.9.2 tarball release
Reproducing:
```bash
yum -y install cmake golang git mercurial
mkdir build
tar -xzf /vagrant/heka-0.9.2.tar.gz -C /root/build/
cd build/heka-0.9.2/
. build.sh
```

Failure:
```
[root@build7 heka-0.9.2]# . build.sh
-- The C compiler identification is GNU 4.8.3
-- Check for working C compiler: /bin/cc
-- Check for working C compiler: /bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Found Go: /bin/go (found suitable version "1.4.2", minimum required is "1.3")
-- Found Git: /bin/git (found version "1.8.3.1")
-- GeoIP.h was not found, GeoIP functionality will not be included in this build.
-- Docker plugins enabled.
-- sphinx-build was not found, the documentation will not be generated.
-- Found Hg: /bin/hg (found version "version 2.6.2")
fatal: Not a git repository (or any of the parent directories): .git
CMake Error at CMakeLists.txt:108 (message):
  Failed to init Heka submodules


-- Configuring incomplete, errors occurred!
make: *** No targets specified and no makefile found.  Stop.
```

Where the failure is `CMakeLists.txt` +108:
```
execute_process(COMMAND "${GIT_EXECUTABLE}" submodule update --init --recursive
WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}"
RESULT_VARIABLE error_code
)
if(error_code)
    message(FATAL_ERROR "Failed to init Heka submodules")
endif()
```

The only submodule is the documentation theme, so this should be wrapped in `if (INCLUDE_DOCUMENTATION)` ... `endif()`

Pull request inbound shortly.

jetpks:
Added `if (INCLUDE_DOCUMENTATION)` block around the `submodule init`
in `CMakeLists.txt`.

fixes #1554

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,12 +100,14 @@ include(CTest)
 include(externals)
 include(mocks)
 
-execute_process(COMMAND "${GIT_EXECUTABLE}" submodule update --init --recursive
-WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}"
-RESULT_VARIABLE error_code
-)
-if(error_code)
-    message(FATAL_ERROR "Failed to init Heka submodules")
+if (INCLUDE_DOCUMENTATION)
+    execute_process(COMMAND "${GIT_EXECUTABLE}" submodule update --init --recursive
+    WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}"
+    RESULT_VARIABLE error_code
+    )
+    if(error_code)
+        message(FATAL_ERROR "Failed to init Heka submodules")
+    endif()
 endif()
 
 configure_file("${CMAKE_SOURCE_DIR}/sandbox/lua/lua_sandbox.go.in" "${HEKA_PATH}/sandbox/lua/lua_sandbox.go" @ONLY)


trink:
The submodule init should not fail, whether you are building documentation or not.  This just masks the issue.

jetpks:
The `.git` directory is not packaged in the tarball release, so there is literally no way for `. build.sh` to ever run unless this is patched in.

trink:
Yeah, but there is also no way to build the documentation from the tarball either (if you attempt to you will still fail with your issue)

rafrombrc:
@jetpks  I assume you're talking about the "Source code (tar.gz)" link that is on each release page, is that right?

Unfortunately, we don't create that link, nor do we want it to be there. Github generates it automatically, along with the similar "Source code (zip)" link that is also there. If there was a way I could remove it from the releases page, I would. We have no intention nor any desire to do source tarball releases.

We *do* provide 32- and 64-bit binary tarball releases for Linux. Those include the documentation, but no source code.

If you want to build from source you should clone the repository from git. That's the only supported build mechanism. I'm sorry the source tarball thing has misled you, but AFAIK there's nothing we can do to get rid of it. If you know any different, or if I'm misunderstanding your issue, please let me know.

jetpks:
Ahhhhh @rafrombrc that makes a lot of sense. I got the tarball from [here](https://github.com/mozilla-services/heka/releases). I ran into this building rpms where source traditionally comes in a tarball. I'll just patch this out at build time. Thanks for the info! :)

PR #1558, opened 2015-06-02
Support for path in elasticsearch server config

xrl:
ElasticSearchOutput's server configuration uses the spec, host,
  and port (e.g. "http://localhost:9200") but ignores the path.
  "http://localhost/es/" becomes "http://localhost/". The path value was
  already in the structure parsed by the go url lib and needed to be tracked
  and added to the HTTP request URL.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -87,6 +87,7 @@ Bug Handling
 
 * Added ability to reconnect after lost connection in `heka-flood` (#1536).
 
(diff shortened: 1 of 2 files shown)


rafrombrc:
This looks great, just needs an entry in the "bug fixes" section for v0.10.0 in the CHANGES.txt changelog. Thanks!

xrl:
Bump with the changelog addition.

PR #1563, opened 2015-06-08
Change delete_idle_stats default value from false to true

ziegeer:
Heka has a delete_idle_stats option that is responsible for removing unused statistics.  The default value is false, which means once Heka has seen a stat, it will repeatedly send it forever regardless of weather it sees that stat come in again.  I propose a default value of true, which is far less likely to cause harm.  In practice, stat names change regularly and I've seen Heka routinely cause more than 500,000 unused, worthless stats to be sent to a single Graphite instance every five seconds.  Even a beefy Graphite instance can struggle to keep up with that volume of data, especially since that's not including the actual live data that people want.  Graphite does not include good facilities for identifying where a given stat comes from so identifying Heka as the source of the flood of stats took many man-hours here.  Please save many headaches and remove unused stats by default by setting delete_idle_stats to true out of the box.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -17,6 +17,8 @@ Backwards Incompatibilities
 
 * The 'make deb' target requires fakeroot and debhelper to be installed.
 
+* Switched delete_idle_stats setting default from false to true so unused metrice are not continuously sent forever.
+
 Features
 --------
 
--- a/docs/source/config/inputs/stataccum.rst
+++ b/docs/source/config/inputs/stataccum.rst
@@ -54,7 +54,7 @@ Config:
     Prefix to use for the statsd `numStats` metric. Defaults to "statsd".
 - delete_idle_stats (bool):
     Don't emit values for inactive stats instead of sending 0 or in the case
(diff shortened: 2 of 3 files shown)


rafrombrc:
The "send idle stats" behavior doesn't originate with Heka. My understanding is that most statsd servers behave this way, i.e. once a statistic is seen once, the statsd server will continue to send that statistic forever regardless of whether or not new data arrives. Some statsd servers (including Heka) have added support for omitting idle stats, but generally that is not the default behavior.

In other words, my understanding is that Heka is just matching what is standard, expected behavior for a statsd server. Do you know what the default behavior is for Etsy's statsd server?

ziegeer:
You are correct, statsd also does not delete idle stats by default.  I asked them for a change as well because that also has caused problems here in the past, and they said they are interested but are waiting for a 1.0 release before they change a default config value like that.

rafrombrc:
Right. Our goal is to be able to work as a drop-in replacement for other statsd servers, so we're going to follow their lead on this one, sorry.

PR #1566, opened 2015-06-09
Fix up Dockerfile for local development

mattrobenolt:
* Switch to using the base `golang:1.4` especially considering heka doesn't compile against `1.3.1` anymore.
* Add new/missing dependencies to compile against latest HEAD
* Add libgeoip to enable that functionality
* Remove `NUM_JOBS` because this is broken (not entirely sure why, but seems like a race condition in the build script
* Don't automatically run `build.sh` on creating the container. Since this is local dev, it's more favorable to get a working  environment first, then try to compile, rather than the compilation potentially causing the `docker build` to fail.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,44 +1,32 @@
 # heka_base image
-FROM debian:jessie
+FROM golang:1.4
 
 MAINTAINER Chance Zibolski <chance.zibolski@gmail.com> (@chance)
 
 RUN     apt-get update && \
         apt-get install -yq --no-install-recommends \
         build-essential \
-        bzr \
         ca-certificates \
         cmake \
-        curl \
-        git \
-        golang-goprotobuf-dev\
-        make \
-        mercurial \
+        debhelper \
+        fakeroot \
+        libgeoip-dev \
+        libgeoip1 \
+        golang-goprotobuf-dev \
         patch \
         ruby-dev \
         protobuf-compiler \
         python-sphinx \
         wget
 
-# Install Go 1.3.1
-RUN curl -s https://storage.googleapis.com/golang/go1.3.1.linux-amd64.tar.gz -o /tmp/go.tar.gz && \
-        echo "3af011cc19b21c7180f2604fd85fbc4ddde97143 /tmp/go.tar.gz" | sha1sum -c && \
-        tar -C /usr/local -xzf /tmp/go.tar.gz
-
 WORKDIR /heka
-
-ENV GOROOT /usr/local/go
-ENV PATH $PATH:/usr/local/go/bin:/go/bin
 
 ENV CTEST_OUTPUT_ON_FAILURE 1
 ENV BUILD_DIR   /heka/build
 ENV GOPATH      $BUILD_DIR/heka
 ENV GOBIN       $GOPATH/bin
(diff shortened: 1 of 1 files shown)


jotes:
I think we should create separate ticket to investigate why NUM_JOBS has race condition.
Thx for your work.

rafrombrc:
I'm testing out this branch, but when I use it and then try to run the `build_docker.sh` command it fails with the following output:

```
... lots of elided output ...
Step 10 : COPY . /heka
 ---> Using cache
 ---> b14a45d99230
Successfully built b14a45d99230
Sending build context to Docker daemon 15.36 kB
Sending build context to Docker daemon 
Step 0 : FROM mozilla/heka_base
 ---> b14a45d99230
Step 1 : RUN mkdir -p /heka_docker
 ---> Using cache
 ---> 000d455d6a01
Step 2 : RUN cd /heka/build && make deb
 ---> Running in 3b2ae3a57bea
/bin/sh: 1: cd: can't cd to /heka/build
The command '/bin/sh -c cd /heka/build && make deb' returned a non-zero code: 2
```

Should this be working?

mattrobenolt:
@rafrombrc So I'm not 100% sure what the intentions of this Dockerfile are, but I explicitly removed the last `RUN` which calls `build.sh`. So since this container isn't running `build.sh`, `/heka/build` never gets created.

I modified this container to suit the needs of a developer wanting to contribute to heka and get into a working environment which I can edit and build on my own inside the container. This is similar IMO to how you contribute to docker itself. They provide a dev environment inside a container in which you can compile and run tests, etc, on docker.

So maybe we can move the call for `build.sh` into https://github.com/mozilla-services/heka/blob/dev/docker/Dockerfile ?

rafrombrc:
The original intention of the Docker setup was that it would create two containers. The first was for the build, which would install and build Heka from source, generating a deb package. This container would in turn create a *second* container, for deployment, into which the deb package would be installed.

I'm not rigidly attached to the binaries and the second container getting generated immediately, but I *am* rigidly attached to the documentation matching reality, and that the docs include clear steps that will get users all the way to deployment. Right now the Docker build instrux in the readme (visible here: https://github.com/mozilla-services/heka/tree/dev/docker) still say that running `build_docker.sh` will complete the build and generate both containers, and the shell script assumes that's what it's trying to accomplish. Removing the build.sh call broke build_docker.sh's assumptions, so now running build_docker.sh exits with an error.

I'd be open to any reasonable solution, but I can't merge this until the docs describe how to get to a successful deployment. :)

mattrobenolt:
Totally reasonable. If you're ok with this change, I'll update the rest to fit.

rafrombrc:
Yes, in general I'm fine with it. As long as I can read the Docker readme and successfully use those instructions to get all the way to the deployed Heka container without it feeling terribly burdensome, I'll be happy.

rafrombrc:
Bump. Just wondering if you are indeed planning on finishing this up so the docs and the actual behavior are in alignment, @mattrobenolt . :)

xrl:
I'm going to do some work on this today.

rafrombrc:
This work was further tweaked and merged in #1696. Thanks!

Issue #1549, opened 2015-05-22
Allow the mesage matcher to accept RFC3339 time stamp strings


trink:
`Timestamp >= "2015-04-30T00:00:00Z"`
in addition to the existing numeric value
`Timestamp >=1430352000000000000`

whd:
:+1:

jotes:
Do you want to work on this issue or can i try dig into it?

trink:
@jotes If you are interested in taking a crack at it in Heka that is fine.

PR #1568, opened 2015-06-09
Added support for parsing of timestamps in message matcher for Timestamp field (#1549)

jotes:


message/message_matcher.go
```diff
@@ -98,6 +98,26 @@ func getNumericValue(msg *Message, stmt *Statement) float64 {
 	return 0
 }
 
+func getTimestampValue(msg *Message, stmt *Statement) float64 {
```
trink: The matcher shouldn't have to know anything about this.  The time should always be converted to a numeric value by the grammar and everything should work as it did before.

PR #1570, opened 2015-06-10
kafka: Add `topic_prefix` and escape topic names

mattrobenolt:
These are especially useful when paired with `topic_variable`.
Since the variable is being extracted from a field, we can't guarantee
that the characters being used for the topic name are valid, so
we implement a slightly modified version of stdlib's `QueryEncode`
which sanitizes the name to the range of valid topic chars.

Example:

```toml
[KafkaOutput]
topic_prefix = "heka-"
topic_variable = "Fields[ContainerName]"
```


plugins/kafka/kafka_output.go
```diff
@@ -371,6 +373,53 @@ func (k *KafkaOutput) CleanupForRestart() {
...
+func shouldEscape(c byte) bool {
+	// List of valid kafka topic characters can be found at:
+	// https://github.com/apache/kafka/blob/43b92f8b1ce8140c432edf11b0c842f5fbe04120/core/src/main/scala/kafka/common/Topic.scala#L25
+	//   val legalChars = "[a-zA-Z0-9\\._\\-]"
+	if 'A' <= c && c <= 'Z' || 'a' <= c && c <= 'z' || '0' <= c && c <= '9' {
+		return false
+	}
+
+	switch c {
+	case '.', '_', '-':
```
mattrobenolt: So there is one problem here, is the `-` character needs to be escaped, since that's being used as the escape character above on L394.

This will prevent backwards compatibility though if we escape `-`. So unsure how to proceed best without changing the topic names on people.

rafrombrc: Not sure I understand. Can't we use a different character for escaping? Back-slashes aren't allowed in a topic name, but they could still be used as an escape character, right? I might be missing something, though...

rafrombrc:
Sorry to take so long to get to this... things have been a bit crazy for me lately. Thanks for your contribution!

Anyway, I didn't love the two mutually exclusive topic options in the first place, and this feels like it goes a bit farther in the wrong direction. I'd rather see all of the settings reduced down to a single `topic` setting, where that setting supports variable interpolation. So your example might look like `topic = "heka-%{Fields[ContainerName]}"`, or even `topic = "heka-%{ContainerName}"`, where the var name is assumed to come from the dynamic fields if it's not one of the base schema field names.

Also, any changes to plugin configuration settings need to be reflected in both the documentation and the CHANGES.txt changelog.

Thanks!

Issue #1561, opened 2015-06-05
ProcessInput silently stops Processing Input


hynd:
I'm encountering a super-weird bug with ProcessInput's.  Not particularly easy to repro, but i think now i know why...

The symptoms are the Heka ProcessInput plugin gets into a weird state where doesn't appear to be reading input (and doesn't exit), but the spawned cmd stays running.  I see the same behaviour from version 0.9.0 to 0.10/dev.

Not all ProcessInput-fired cmd's trigger the behaviour, just those that have particular output characteristics - those that end up splitting the output into multiple `write(2)` calls where the record content before a delimiter is broken across them.  (Python buffers it's stdout by default - it's possible to manipulate that in various ways, but in some cases the "\n" is still sent in a separate `write()`, triggering this bug).

The following Heka .toml and Python generates batches of data >4KB (which should get split into multiple write()s) hopefully reproduces the problem...  I say "hopefully", because this only seems to happen on boxes that are busy (my dev/test boxes don't hit it so frequently) - sometimes the ProcessInput stops producing data immediately, sometimes after just a couple of `ticker_intervals`.  Turns out strace'ing the Python also seemed a fairly consistent way to trigger it.  I guess both the extra load or strace increases the delay between write()s...

```toml
[ProcessInput]
stdout = true
stderr = false
splitter = "TokenSplitter"
ticker_interval = 0

[ProcessInput.command.0]
bin = "/path/to/test.py"

[PayloadEncoder]
append_newlines = false

[LogOutput]
encoder = "PayloadEncoder"
message_matcher = "TRUE"
```

```python
#!/usr/bin/python
import sys
import time

def main():
    while True:
        for i in xrange(50):
          print "x"*128
        sys.stdout.flush()
        time.sleep(5)

if __name__ == "__main__":
    main()
```

[...]

PR #1572, opened 2015-06-13
Stop zero-length records with no error ending the splitter (#1561)

hynd:
Only if there's more data to read and there's no error.
Continue rather than attempting to deliver an empty record.


matschaffer:
Left https://gist.github.com/matschaffer/55864a96702f72904087 running over the weekend with this patch and it's still processing records this morning. :+1:

PR #1569, opened 2015-06-10
DockerLogInput: add new `fields_from_env`

mattrobenolt:
Example:

```toml
[DockerLogInput]
fields_from_env = [ "FOO", "BAR" ]
```

PR #1574, opened 2015-06-18
Update Schema InfluxDB Write Encoder to use InfluxDB 0.9.x+ line protocol

acesaro:
This PR updates the Schema InfluxDB Write Encoder to format metrics into the line protocol instead of the JSON format that was originally proposed for the 0.9.0 release.  This also makes sure that proper formatting is done for the various fields to escape spaces, commas and double quotes as defined by the line protocol.  There is also a somewhat hacky implementation to overcome the fact that Lua converts float values of 0.0 into an integer based on its internal representation of numerical data types.  I've also added a couple of new configuration items that provide more flexibility in the naming of measurements as they are sent to InfluxDB.  There is now more emphasis on utilizing tags instead of the Graphite style "paths" to uniquely identify series, so the defaults have been blanked out to work with this recommendation more naturally.

### DIFF
--- a/sandbox/lua/encoders/schema_influx_write.lua
+++ b/sandbox/lua/encoders/schema_influx_write.lua
@@ -3,16 +3,14 @@
 -- file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 --[=[
-Converts full Heka message contents to JSON for InfluxDB HTTP write API (new
-in InfluxDB v0.9.0). Optionally includes all standard message fields as tags
-or fields and iterates through all of the dynamic fields to add as points
-(series), skipping any fields explicitly omitted using the `skip_fields`
-config option.  It can also map any Heka message fields as tags in the request
-sent to the InfluxDB API, using the `tag_fields` config option. Tags,
-timestamp/precision, database, and retention policy are defined once for all
-points submitted in the API call.  All dynamic fields in the Heka message are
-converted to separate series in the `points` array that is submitted to
-InfluxDB.
+Converts full Heka message contents to line protocol for InfluxDB HTTP write API
+(new in InfluxDB v0.9.0).  Optionally includes all standard message fields
+as tags or fields and iterates through all of the dynamic fields to add as
+points (series), skipping any fields explicitly omitted using the `skip_fields`
+config option.  It can also map any Heka message fields as tags in the
+request sent to the InfluxDB write API, using the `tag_fields` config option.
+All dynamic fields in the Heka message are converted to separate points
+separated by newlines that are submitted to InfluxDB.
 
 .. note::
     This encoder is intended for use with InfluxDB versions 0.9 or greater. If
@@ -21,8 +19,8 @@ InfluxDB.
 
 Config:
 
-- name_prefix (string, optional, default "name")
-    String to use as the `name` key's prefix value in the generated JSON.
+- name_prefix (string, optional, default "")
+    String to use as the `name` key's prefix value in the generated line.
     Supports :ref:`message field interpolation<sandbox_msg_interpolate_module>`.
     `%{fieldname}`. Any `fieldname` values of "Type", "Payload", "Hostname",
     "Pid", "Logger", "Severity", or "EnvVersion" will be extracted from the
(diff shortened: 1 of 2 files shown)


davidbirdsong:
@acesaro mind pointing out the area that covers the Lua float to int problem? I've been working on an encoder that converts 0.9 json style to line protocol, but I've been dead in the water on that issue for some time now.

sandbox/lua/encoders/schema_influx_write.lua
```diff
@@ -281,27 +277,11 @@ function process_message()
...
-            }
-
-            -- Increment the index to the table so the next iteration will
-            -- append another entry to the array. This allows us to send a
-            -- bunch of metrics to InfluxDB with a single HTTP request
-            points_index = points_index + 1
+        -- Force to remain as a true float if value is 0; otherwise
+        -- InfluxDB will reject the point if previous values were
+        -- actual floats
+        if value == 0 and typ == 3 then
```
acesaro: @davidbirdsong It's pretty ugly, but the workaround for lua converting 0.0 to 0 is here and then followed up on line 334.  It basically just converts it to a string and then removes the quotes when it's concatenating values when injecting a new message with payload composed of one or more series that are newline separated.

acesaro:
I started thinking about this a bit more and this encoder is very similar to functionality that I've been wanting to build for Graphite formatted metrics to send to Carbon.  I've actually already implemented it by adding a boolean config variable `carbon_line_format` and some if statements to just format the same data differently (and skip tag generation steps that aren't relevant).

What are some thoughts about shifting the focus of this encoder to a common InfluxDB/Carbon line protocol encoder rather than copying most of the script and trimming it down for Carbon only?  Perhaps move some of the common functionality int a lua module instead and then have the two encoders reference common functions with that and be trimmed down to the essentials within them?

rafrombrc:
I'm a very strong +1 to moving the shared functionality into a module and having a separate encoder for each endpoint. Sounds like a great idea. I haven't looked at your PR yet (am just now surfacing after Monitorama and traveling home) so I don't know what's going on in there, but I should point out that I've already created a `graphite` module with some utility functions to help generate graphite formatted stats from statsd-like input, which may be useful in some way: https://github.com/mozilla-services/heka/blob/dev/sandbox/lua/modules/graphite.lua

acesaro:
Cool, thanks for the feedback.  I'm going to go ahead and close this PR and come up with a new one.  I did take a look at the graphite module that you mentioned above (and at Monitorama) and I think I'm going to need a little bit of guidance on how it could be utilized.  I'll also take a closer look at the gist that you created that shows it in use.  We may want to rename the module to be something more generic since it will be used for different encoders eventually.

haf:
As a newbie: does heka (https://registry.hub.docker.com/u/ianneub/heka/tags/manage/) latest support 0.9 of influxdb or not?

acesaro:
With the JSON protocol, yes it (heka dev branch) does.  That is eventually being deprecated by the InfluxDB project, though, which is the reason that we need to transition to the line protocol which I am working on implementing.

haf:
Thanks.

When is the dev branch stable?

haf:
For the linked docker, heka cannot push to influxdb.

Does this docker container contain enough of the dev branch? https://github.com/disqus/docker-heka

Could you give me a link to the JSON protocol configuration?

acesaro:
That Dockerfile points to someone's fork of the mozilla-services/heka git repository, so I don't know what it would be merged with.  Here's the documentation on the Heka Schema InfluxDB Write Encoder:  https://hekad.readthedocs.org/en/latest/config/encoders/schema_influx_write.html

haf:
It points to @matrobenolt's fork which you have in your unmerged PRs https://github.com/mozilla-services/heka/pull/1569 -- it seems to be the latest. At least when sorting on updated-date on docker. Do you have an official one?

mattrobenolt:
Our fork is just master + a few of my PRs applied that we need in production.

rafrombrc:
Actually it's likely dev and not master (dev is the default branch, and where most new development takes place).

I've given feedback on all 3 of @mattrobenolt 's PRs (thanks for your contributions!) and hope to get them merged very soon.

PR #1585, opened 2015-06-25
Added support for parsing of timestamps in message matcher for Timestamp field (#1549)

jotes:
@rafrombrc @trink  could you review? thanks in advance.

### DIFF
--- a/docs/source/message_matcher.rst
+++ b/docs/source/message_matcher.rst
@@ -67,6 +67,9 @@ Message Variables
     - **Timestamp**
     - **Severity**
     - **Pid**
+- DateTime
+    - **Timestamp**
+    - Message matcher supports matching of dates in RFC3339 format, e.g. '2014-02-03T14:02:03Z'
 - Fields
     - **Fields[_field_name_]** (shorthand for Field[_field_name_][0][0])
     - **Fields[_field_name_][_field_index_]** (shorthand for Field[_field_name_][_field_index_][0])
--- a/message/message_matcher.go
+++ b/message/message_matcher.go
@@ -14,7 +14,9 @@
 
 package message
 
+import "fmt"
 import "strings"
+import "time"
 
 // MatcherSpecification used by the message router to distribute messages
 type MatcherSpecification struct {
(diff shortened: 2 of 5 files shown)


docs/source/message_matcher.rst
```diff
@@ -67,6 +67,9 @@ Message Variables
     - **Timestamp**
     - **Severity**
     - **Pid**
+- DateTime
```
trink: We don't actually have a DateTime data type.  I would just update the existing entry to:

**Timestamp** (Nanoseconds since the Unix epoch or a quoted RFC3339 string that will be converted to nanoseconds e.g., '2014-02-03T14:02:03Z')

message/message_matcher.go
```diff
@@ -243,3 +245,13 @@ func testExpr(msg *Message, stmt *Statement) bool {
 	}
 	return false
 }
+
+func parseFullDateTime(timestamp string) (tdata time.Time, err error) {
```
trink: This helper function should go in the .y file since it is only used there

message/message_matcher_parser.y
```diff
@@ -130,12 +129,30 @@ string_test : string_vars relational STRING_VALUE
        //fmt.Println("string_test regexp", $1, $2, $3)
        nodes = append(nodes, &tree{stmt:&Statement{$1, $2, $3}})
        }
+   |   VAR_TIMESTAMP relational STRING_VALUE
+       {
+       //fmt.Println("timestamp_string_test regexp", $1, $2, $3)
```
trink: copy and paste error in commented out debug message 'regexp'

message/message_matcher_parser.y
```diff
@@ -130,12 +129,30 @@ string_test : string_vars relational STRING_VALUE
...
 numeric_test : numeric_vars relational NUMERIC_VALUE
-   {
-   //fmt.Println("numeric_test", $1, $2, $3)
-   nodes = append(nodes, &tree{stmt:&Statement{$1, $2, $3}})
-   }
+        {
+        //fmt.Println("numeric_test", $1, $2, $3)
+        nodes = append(nodes, &tree{stmt:&Statement{$1, $2, $3}})
+         }
+   |    VAR_TIMESTAMP relational NUMERIC_VALUE
```
trink: You should be able to leave VAR_TIMESTAMP in numeric_vars and get rid of this grammar (the string case is special cased and tested first)

PR #1593, opened 2015-06-30
Use correct Apache combined format

sathieu:


sathieu:
%b is used instead of %O.

sandbox/lua/decoders/apache_access.lua
```diff
@@ -50,13 +50,16 @@ Config:
     type = "combined"
     user_agent_transform = true
     # combined log format
-    log_format = '%h %l %u %t \"%r\" %>s %O \"%{Referer}i\" \"%{User-Agent}i\"'
+    log_format = '%h %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-agent}i\"'
+
+    # combined I/O log format
+    log_format = '%h %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-agent}i\" %I %O'
```
trink: Comment out this line in the example

trink:
Also if you look at the actual config installed with a default apache the existing values are correct.

PR #1406, opened 2015-03-10
Always encode into MsgBytes before injection into router.

rafrombrc:

PR #1595, opened 2015-07-02
Replace InfluxDB Write Encoder with InfluxDB Line Encoder and add Carbon Line Encoder

acesaro:
This PR removes a deprecated encoder with an updated one that simplifies its implementation by moving common functionality into new modules, field_util and line_protocol.  As a result, I was able to easily implement the Schema Carbon Line Encoder, which is very similar to the Schema InfluxDB Line Encoder, just without tags and named value fields.  These modules can be used for any other similar endpoint systems that we wish to integrate with that use a similar "line protocol" interface.


sandbox/lua/modules/field_util.lua
```diff
@@ -0,0 +1,173 @@
...
+    local interp_field
+    -- If the %{field} substitutions are defined in the field,
+    -- replace them with the actual values from the message here
+    if string.find(field, "%%{[%w%p]-}") then
+        interp_field = interp.interpolate_from_msg(field, nil)
+    else
+        interp_field = field
+    end
+    return interp_field
+end
```
rafrombrc: This function doesn't seem to add much to the `interpolate_from_msg` function to which it delegates. The `string.find` check might save a few cycles in cases where there's nothing to interpolate, but even if that's worthwhile it might be better off just being added to the original function.

Am I missing something re: why it makes sense to have a wrapper function?

sandbox/lua/modules/field_util.lua
```diff
@@ -0,0 +1,173 @@
...
+            if field == "**all_base**" then
+                all_base_fields = true
+            end
+            if field == "**all**" then
+                all_fields = true
+            end
+        end
+
+        for field in pairs(base_fields_list) do
+            if all_base_fields or all_fields then
```
rafrombrc: Much more efficient to check the `if`  outside the loop rather than inside.

sandbox/lua/modules/line_protocol.lua
```diff
@@ -0,0 +1,207 @@
...
+
+**influxdb_line_msg()**
+    Wrapper function that calls others within this module and the field_util
+    module to generate a table of InfluxDB line protocol messages that are
+    derived from the base or dynamic fields in a Heka message. Base fields or
+    dynamic fields can be used in the metric name portion of the message and
+    included as tags if desired.  Configuration is implemented in the encoders
+    that utilize this module.  This function takes no arguments.
+
+--]=]
```
rafrombrc: Seems like this should probably document the attribute portion of the API, since AFAICT this module won't really work correctly unless you set the attribute values that it expects. Also it would be nice to explicitly define the settable variables in the code below, with a comment explaining that the values are expected to be mutated by calling code prior to any use of the API functions.

sandbox/lua/modules/line_protocol.lua
```diff
@@ -0,0 +1,207 @@
...
+if not name_prefix_delimiter then
+    name_prefix_delimiter = ""
+end
+if not value_field_name then
+    value_field_name = "value"
+end
+
+-- Remove blacklisted fields from the set of base fields that we use, and
+-- create a table of dynamic fields to skip.
+local skip_fields_str = read_config("skip_fields") or ""
```
rafrombrc: Feels a bit strange to me that most of the config values are pulled out in the encoder code, but there are a few oddballs that are extracted by the module code. Maybe these should be turned into a part of the attribute API for consistency?

sandbox/lua/modules/line_protocol.lua
```diff
@@ -0,0 +1,207 @@
...
+local skip_fields, skip_fields_all_base = field_util.field_map(skip_fields_str)
+local used_base_fields = field_util.used_base_fields(skip_fields)
+
+-- Create and populate a table of fields to be used as tags
+local tag_fields_str = read_config("tag_fields") or "**all_base**"
```
rafrombrc: Same as above re: `read_config`.

rafrombrc: Okay, I think I see why you did it this way, since you're using these values in the `used_base_fields` and `field_map` calls happening at module scope, and you need to make sure you have the config values even before they might be set by the client code. I'm still a bit wary of it, though.

Maybe instead of having the client code set attributes on the `line_protocol` module, we can have a `line_protocol_config` function that the client code must call before calling any other functions in the module, passing the needed values in as separate arguments, or maybe embedded in a Lua table. That smells a bit better than setting module-level attributes from outside, even if internally that's still what we're doing. It would also mean that all of the `read_config` calls can be moved to the encoders.

sandbox/lua/modules/field_util.lua
```diff
@@ -0,0 +1,173 @@
...
+base_fields_list = {
+    Type = true,
+    Payload = true,
+    Hostname = true,
+    Pid = true,
+    Logger = true,
+    Severity = true,
+    EnvVersion = true
+}
+
```
rafrombrc: Shouldn't this include `Logger`?

sandbox/lua/modules/line_protocol.lua
```diff
@@ -0,0 +1,207 @@
...
+    for field in pairs(field_util.base_fields_list) do
+        local base_field_value = read_message(field)
+        field = field:gsub("([ ,])", "\\%1")
+        if (tag_fields_all or tag_fields_all_base)
+            and field_util.base_fields_tag_list[field] and type(base_field_value) ~= "number" then
+            table.insert(message_tags, field.."="..tostring(base_field_value))
+        elseif used_tag_fields[field] and field_util.base_fields_tag_list[field]
+            and type(base_field_value) ~= "number" then
+            table.insert(message_tags, field.."="..tostring(base_field_value))
+        end
```
rafrombrc: It might involve a small amount of code duplication, but I think this could be made more efficient by doing the `tag_fields_all or tags_fields_all_base` check outside of the loop. If the check is true, then you iterate through the entire base_fields_list like you're already doing. If false, then you only need to iterate through the base_fields_tag_list. Does that seem right to you?

rafrombrc:
Okay, first pass through the code is finished, all comments made inline. Thanks so much for tackling this!

rafrombrc:
P.S. Make sure you read through all of my comments before tackling the changes suggested in any of them. I made a suggestion later (in the "Okay, I think I see why you..." comment) that supersedes the earlier ones, would hate for you to not notice that and do work that you would then need to immediately undo. :P

sandbox/lua/modules/field_util.lua
acesaro: It doesn't add much. I'll migrate the if check to the msg_interpolate module and add it to this PR, updating references to field_interp to point to msg_interpolate.interpolate_from_msg instead.

sandbox/lua/modules/field_util.lua
acesaro: Hmm maybe I am missing what you're pointing out.  Logger is part of both of these tables (base_fields_list and base_fields_tag_list).

sandbox/lua/modules/line_protocol.lua
acesaro: Ya, this is really weird what I've done here, the same action is taken in both if conditions.  I like your idea and will make updates accordingly.  Thanks.

acesaro: I actually cleaned this up and moved the functionality into the `points_tags_tables` function, since it really only needs to loop through `base_fields_tag_list` (the whole point of that table even existing).  I kept the for loop inside a conditional that checks if we are working with `carbon_format` so it will never run for Carbon but with always run for InfluxDB if there is configuration in place to add tags.  It then only matches if any of the previously checked conditions is true.  There is a little code duplication, but net result is improved, IMO.
(updates to the PR will be made later after this comment is posted).

acesaro:
I've made some major improvements based on your excellent suggestions.  :)  Please give it another review when you have a chance.  I'll add some commentary around some of the key areas that have been updated.  Moving the configuration of the line_protocol functions to that module was a great improvement, so thanks for that and pointing out some of the other inefficiencies.  I'm learning a lot!

sandbox/lua/encoders/schema_carbon_line.lua
```diff
@@ -0,0 +1,120 @@
...
+- skip_fields (string, optional, default nil)
+    Space delimited set of fields that should *not* be included in the
+    Carbon records being generated. Any fieldname values of "Type",
+    "Payload", "Hostname", "Pid", "Logger", "Severity", or "EnvVersion" will
+    be assumed to refer to the corresponding field from the base message
+    schema. Any other values will be assumed to refer to a dynamic message
+    field. The magic value "**all_base**" can be used to exclude base fields
+    from being mapped to the event altogether.
+
+- source_value_field (string, optional, default nil)
```
acesaro: I added this config option and associated functionality since I'm primarily using this encoder to take the output of data collection scripts, splitting messages from it by newline, parsing them with the Payload Regex Decoder and then parsing them to Carbon lines with this encoder.  This allows the encoder to work the original way (by sending multiple fields as separate Carbon lines) or the way I just described, which identifies a single field in the message that contains the value (e.g. "5.0") and adds it to the Carbon line.

sandbox/lua/encoders/schema_carbon_line.lua
```diff
@@ -0,0 +1,120 @@
...
+    dev.myhost.stats.loadavg.5MinAvg 0.11 1434932023
+
+--]=]
+
+local line_protocol = require "line_protocol"
+
+local decoder_config = {
+    carbon_format = true,
+    timestamp_precision = "s",
+    name_prefix = read_config("name_prefix") or nil,
```
acesaro: Let me know if you think a value like `false` would be better suited than `nil` for a default.  Right now the code checks the value of these config options and works accordingly, so `nil` is working fine, but I just wanted to get your opinion based on past experience.

sandbox/lua/encoders/schema_influx_line.lua
```diff
@@ -0,0 +1,165 @@
...
+    sent to InfluxDB as measurements, but still include them as tags.
+
+- timestamp_precision (string, optional, default "ms")
+    Specify the timestamp precision that you want the event sent with.  The
+    default is to use milliseconds by dividing the Heka message timestamp
+    by 1e6, but this math can be altered by specifying one of the precision
+    values supported by the InfluxDB write API (ms, s, m, h). Other precisions
+    supported by InfluxDB of n and u are not yet supported.
+
+- value_field_key (string, optional, default "value")
```
acesaro: I decided to change the name of this field based on InfluxDB vocabulary in their documentation.

sandbox/lua/modules/line_protocol.lua
```diff
@@ -0,0 +1,246 @@
...
+local pairs = pairs
+local string = require "string"
+local table = require "table"
+local tostring = tostring
+local type = type
+
+local M = {}
+setfenv(1, M) -- Remove external access to contain everything in the module.
+
+local function influxdb_kv_fmt(string)
```
acesaro: I broke this out into its own function since the substitution was made in more than two different spots in the code.

sandbox/lua/modules/line_protocol.lua
```diff
@@ -0,0 +1,246 @@
...
+
+    -- Initialize the tags table, including base field tag values in
+    -- list if the magic **all** or **all_base** config values are
+    -- defined.
+    local tags = {}
+    if not config["carbon_format"]
+        and (config["tag_fields_all"]
+        or config["tag_fields_all_base"]
+        or config["used_tag_fields"]) then
+            for field in pairs(field_util.base_fields_tag_list) do
```
acesaro: Removed the separate function for this and moved it back into this function, simplifying the implementation based on what it actually needed to achieve (looping through "taggable" base fields, as defined by the base_fields_tag_list table.

sandbox/lua/modules/line_protocol.lua
```diff
@@ -0,0 +1,246 @@
...
+        -- Include the dynamic fields as tags if they are defined in
+        -- configuration or the magic value "**all**" is defined.
+        -- Convert value to a string as this is required by the API
+        if not config["carbon_format"]
+            and (config["tag_fields_all"]
+            or (config["used_tag_fields"] and used_tag_fields[name])) then
+                table.insert(tags, influxdb_kv_fmt(name).."="..tostring(influxdb_kv_fmt(value)))
+        end
+
+        if config["source_value_field"] and name == config["source_value_field"] then
```
acesaro: This was changed from using `exclude_name` to the config value that replaced it (and makes it more flexible), `source_value_field`, which indicates which field in the message to pull the value from.

sandbox/lua/modules/line_protocol.lua
```diff
@@ -0,0 +1,246 @@
...
+    -- the base fields serve as tags for them. If skip_fields does not
+    -- define any base fields, they are added to the fields of each data
+    -- point and each dynamic field value is set as the "value" field.
+    -- Setting skip_fields to "**all_base**" is recommended to avoid
+    -- redundant data being stored in each data point (the base fields
+    -- as fields and as tags).
+    local points = {}
+    while true do
+        -- Iterate through Fields array in the message
+        local typ, name, value = read_next_field()
```
trink: If you are up to it this should be changed to read_message("raw")/decode_message/iterate.  If not we will update it when the read_next_field API is removed.  https://github.com/mozilla-services/heka/issues/1602

sandbox/lua/modules/line_protocol.lua
```diff
@@ -0,0 +1,246 @@
...
+API
+^^^
+
+**carbon_line_msg(config)**
+    Wrapper function that calls others within this module and the field_util
+    module to generate a table of Carbon line protocol messages that are
+    derived from the dynamic fields in a Heka message. Base fields or
+    dynamic fields can be used in the metric name portion of the message.
+    Configuration is implemented in the encoders that utilize this module.
+    This function takes no arguments.
```
rafrombrc: That last sentence is no longer true. :)

sandbox/lua/modules/line_protocol.lua
```diff
@@ -0,0 +1,246 @@
...
+        a Carbon server after being looped through in an encoder implementing
+        this API.
+
+**influxdb_line_msg(config)**
+    Wrapper function that calls others within this module and the field_util
+    module to generate a table of InfluxDB line protocol messages that are
+    derived from the base or dynamic fields in a Heka message. Base fields or
+    dynamic fields can be used in the metric name portion of the message and
+    included as tags if desired.  Configuration is implemented in the encoders
+    that utilize this module.  This function takes no arguments.
```
rafrombrc: Same re: the last sentence.

sandbox/lua/modules/line_protocol.lua
```diff
@@ -0,0 +1,246 @@
...
+            of this API.  Defaults for this module are defined within
+            the set_config function and clients implementing this API
+            can reference it for available options.
+
+    *Return*
+        A table of InfluxDB line protocol messages ready to be sent to
+        an InfluxDB server after being looped through in an encoder
+        implementing this API.
+
+--]=]
```
rafrombrc: Need to add documentation for the `set_config` function.

sandbox/lua/modules/line_protocol.lua
acesaro: Thanks @trink.  I am happy to do so if you have any examples of how I would replace `read_next_field` with the functionality you described.  In the meantime, I'll parse through the existing lua code to see if I can find any use of that.

acesaro: Well, I tried to implement it similar to below, but `read_message("raw")` is returning nil in the place that `read_next_field` was used.  Any suggestions?

```lua
    local decoded_message = decode_message(read_message("raw"))
    for field, value in pairs(decoded_message) do
...
        if config["carbon_format"] then
            field = field:gsub("[^%w]", "_")
        end
...
```

cablehead: Just came across this as I'm trying to get read_message("raw") to work. It's returning nil for me, inside a lua sandbox encoder plugin as well...

Any advice would be appreciated!

trink: In this case you want to iterate over the array of message fields. As per the decode_message documentation it returns:
`The array based version of the message structure with the value member always being an array (even if there is only a single item). This format makes working with the output more consistent...`

 This code works correctly for me in an encoder on the dev branch.
```lua
function process_message()
    local msg = decode_message(read_message("raw"))
    if not msg.Fields then return 0 end

    for i, v in ipairs(msg.Fields) do
        add_to_payload(v.name, " =")
        for m, n in ipairs(v.value) do
            add_to_payload(" ", n)
        end
        add_to_payload("\n")
    end
    inject_payload("txt", "fields")
    return 0
end
```

sandbox/lua/modules/line_protocol.lua
```diff
@@ -0,0 +1,246 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
```
trink: I would also rename this module 'line_protocol' is too generic for and influxDB specific module.

rafrombrc: The module supports Graphite stat format in addition to InfluxDB... maybe 'stat_formatting' or something similar?

acesaro: How about 'ts_line_protocol' since it's being used for different time series databases (and could be expanded to others that implement a 'line protocol' in the future)?

rafrombrc: Sounds great. And thanks so much for staying with us through all of the feedback. :)

sandbox/lua/modules/line_protocol.lua
acesaro: Hmm, I'm still getting a null result from `read_message("raw")`, resulting in this error message:
```
2015/07/04 00:15:45 Plugin 'LogOutput' error: Error encoding message: FATAL: process_message() 
.../local/heka/share/heka/lua_modules/line_protocol.lua:135:bad argument #0 to 'decode_message'
(must have one string argument)
```
I'm using the 0.9.2 release for testing these modules/encoders.  Has this functionality been updated since that release?  I can try with the dev branch and report back with my findings.

rafrombrc: Yes, dev branch will help you here. The `read_message("raw")` call technically works on 0.9.2, but it doesn't give you the desired result b/c the `pack.MsgBytes` value (which is what `read_message("raw")` is accessing) isn't guaranteed to have the protobuf encoding for the message. This changed when #1406 was merged to dev, which ensures that once a pack hits the router, the protobuf encoding will always be available.

acesaro:
Ok, I got the iteration of the fields in a message moved to using `decode_message(read_message("raw"))` after testing with a build of hekad from the latest dev branch commit and updated this PR.  I believe I've also addressed the other outstanding requests.  Please git it another review at your earliest convenience and let me know of anything else you find.  Thanks!

acesaro:
Closing this PR in lieu of another that will be merged into the versions.0.10 branch.

PR #1596, opened 2015-07-02
ESJsonEncoder support for selecting subset of Fields[]

xrl:
I have been using the Fields structure to store information for my sandbox plugins and I needed to select just one field for output. The current approach is all or nothing, you can choose to send all "Fields" values or you can choose to send none. I add a new configuration value called "DynamicFields" which lets you specify properties, which if present, will be put in the JSON object. This also support raw encoding of DynamicFields values through "DynamicRawBytesFields".

This is meeting my needs right now so I thought I'd share. I'm happy to update the documentation if you want to accept this patch.


rafrombrc:
I'm interested in accepting this, yes, but I'd want to see a few changes to the behaviour first:

* The already existing `fields` setting should take precedence. That is, if `"Fields"` is not included in the original setting's list, then no dynamic fields will be included no matter what the `dynamic_fields` value is. If `"Fields"` *is* included in the original setting and `dynamic_fields` is empty, then all of the dynamic fields will be included. If `"Fields"` is included in the list and `dynamic_fields` is not empty, then only those listed in `dynamic_fields` will be included. Hopefully that makes sense.

* There's no need for `DynamicRawBytesFields`, b/c the `raw_bytes_fields` setting already only applies to dynamic fields. The behavior of the `raw_bytes_field` won't change; if a field is in the `raw_bytes_field` list and it *is* going to be included based on the other config settings, then it will be included as a raw field. If a field is in the `raw_bytes_field` list but it's not going to be included, then it will have no effect.

* The same behaviour should be added to the ESLogstashV0Encoder.

xrl:
Cool, sounds fair to me. I'll work on that and get back to you.

rafrombrc:
Hey! I'd like to have this land so that it can be in the 0.10.0 beta release, which I'm hoping to get out this week. I'm wondering if you have some time available to a) make the changes requested in my comment above and b) close this PR and reopen another one against the versions/0.10 branch (instead of dev).

If you're too busy to tackle that today or tomorrow, no problem, just let me know and I'll take care of making the changes and getting it merged to the right place.

Thanks!

xrl:
I'll work on it first thing tomorrow and push up the changes. Thanks!

rafrombrc:
Awesome, thanks! Don't forget to move this work to a new branch cut from versions/0.10. That also means you'll have to close this PR and reopen a new one that merges back to versions/0.10 instead of dev. :)

xrl:
Yep! Consider it closed :sunglasses:

PR #1609, opened 2015-07-09
Es dynamic fields

xrl:
This pull request addresses feedback from #1596.

I have been using the Fields structure to store information for my sandbox plugins and I needed to select just one field for output. The current approach is all or nothing, you can choose to send all "Fields" values or you can choose to send none. I add a new configuration value called "DynamicFields" which lets you specify properties, which if present, will be put in the JSON object. This also support raw encoding of DynamicFields values through "RawBytesFields".

Do not merge: does not yet support the ESLogstashV0Encoder.


xrl:
Ehh, I don't think this is quite right it's bringing in changes from dev.

PR #1610, opened 2015-07-09
ES dynamic_fields

xrl:
This pull request addresses feedback from #1596.

I have been using the Fields structure to store information for my sandbox plugins and I needed to select just one field for output. The current approach is all or nothing, you can choose to send all "Fields" values or you can choose to send none. I add a new configuration value called "DynamicFields" which lets you specify properties, which if present, will be put in the JSON object. This also support raw encoding of DynamicFields values through "RawBytesFields".

Do not merge: does not yet support the ESLogstashV0Encoder.

### DIFF
--- a/docs/source/config/encoders/esjson.rst
+++ b/docs/source/config/encoders/esjson.rst
@@ -36,7 +36,7 @@ Config:
     indexed into ElasticSearch. Available fields to choose are "Uuid",
     "Timestamp", "Type", "Logger", "Severity", "Payload", "EnvVersion", "Pid",
     "Hostname", and "Fields" (where "Fields" causes the inclusion of any and
-    all dynamically specified message fields. Defaults to including all of the
+    all dynamically specified message fields unless dynamic_fields is specified). Defaults to including all of the
     supported message fields.
 - timestamp (string):
     Format to use for timestamps in generated ES documents. Allows to use
@@ -59,6 +59,8 @@ Config:
     Maps Heka message fields to custom ES keys. Can be used to implement a custom format
     in ES or implement Logstash V1. The available fields are "Timestamp", "Uuid",
     "Type", "Logger", "Severity", "Payload", "EnvVersion", "Pid" and "Hostname".
+- dynamic_fields ([]string):
+    This specifies a subset of keys from the dynamic message property to send to ES. Using this option requires setting 'Fields' in the fields configuration.
 
 Example
 
--- a/plugins/elasticsearch/encoders.go
+++ b/plugins/elasticsearch/encoders.go
@@ -10,6 +10,7 @@
 # Contributor(s):
 #   Tanguy Leroux (tlrx.dev@gmail.com)
 #   Rob Miller (rmiller@mozilla.com)
+#   Xavier Lange (xavier.lange@viasat.com)
 #
 # ***** END LICENSE BLOCK *****/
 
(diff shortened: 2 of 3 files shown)


docs/source/config/encoders/esjson.rst
```diff
@@ -36,7 +36,7 @@ Config:
     indexed into ElasticSearch. Available fields to choose are "Uuid",
     "Timestamp", "Type", "Logger", "Severity", "Payload", "EnvVersion", "Pid",
     "Hostname", and "Fields" (where "Fields" causes the inclusion of any and
```
rafrombrc: I think "Fields" should be changed to "DynamicFields", so there's a more clear link btn the value that is here and the `dynamic_fields` setting that it's connected to. I was originally thinking we could leave "Fields" in as a deprecated option for one release cycle, for b/w compatibility, but I'm not sure it's worth it...

plugins/elasticsearch/encoders.go
```diff
@@ -326,24 +350,52 @@ func (e *ESJsonEncoder) Encode(pack *PipelinePack) (output []byte, err error) {
...
 			writeStringField(first, &buf, e.fieldMappings.Hostname, m.GetHostname())
 		case "fields":
-			for _, field := range m.Fields {
-				raw := false
-				if len(e.rawBytesFields) > 0 {
-					for _, raw_field_name := range e.rawBytesFields {
-						if *field.Name == raw_field_name {
-							raw = true
+			if len(e.dynamicFields) > 0 {
+				fmt.Printf("has dynamic fields %s", e.dynamicFields)
```
rafrombrc: Forgot some debug output.

plugins/elasticsearch/encoders.go
```diff
@@ -326,24 +350,52 @@ func (e *ESJsonEncoder) Encode(pack *PipelinePack) (output []byte, err error) {
...
+				for _, field := range m.Fields {
```
rafrombrc: It seems like it might not be uncommon for there to be a large number of fields, with only a few of them being sent to ES. I think it might make more sense, then, to iterate through `e.dynamicFields` and grab all of the field values by name, rather than iterating through all of the fields and then checking to see if each one is in the list.

xrl: How do you grab fields by name in go code?

docs/source/config/encoders/esjson.rst
xrl: I think it's worth it because people's config's won't break when the load this config. Also, how will people choose to send all Fields (the current default) without this behavior?

plugins/elasticsearch/encoders.go
rafrombrc: Ignore me. You fetch fields by name in go code by using [message.FindAllFields](https://github.com/mozilla-services/heka/blob/dev/message/message.go#L501), but I looked at the implementation of that method and realized that it's just iterating through the fields every time you call it, so it doesn't help after all. :P
Given that, though, I wonder if it doesn't make sense to merge both of the loops (i.e. len(`e.dynamicFields > 0` and not) into one. It means an extra 'if' check on each iteration of the loop when there are no dynamic fields, but that's not a lot of cycles, and you can pre-compute `e.dynamicFields > 0` before the loop so you're just checking a boolean value.

docs/source/config/encoders/esjson.rst
rafrombrc: Other than changing "Fields" to "DynamicFields", the behaviour would be identical. So including "DynamicFields" in the list without setting a separate `dynamic_fields` setting would include all of them, just like it does now.

I'm okay w/ breaking the config, as long as we explicitly fail... i.e. Heka should fail to start if "Fields" (or anything other than the set of supported values, actually) is in the list. What I *don't* want to do is have Heka start up and silently do the wrong thing.

PR #1622, opened 2015-07-17
Fixing small typo bug in ts_line_protocol.lua when quote notation was…

acesaro:
… replaced with dot notation for table key lookups.  Also including a bugfix for missing tags from non-base fields.


rafrombrc:
Thanks for cleaning up my clean-up. :) I've also fixed up the documentation links so the docs you wrote will show up correctly on the docs site.

PR #1631, opened 2015-07-23
Make StatAccumInput Plugin Support Multi percent_threshold values

i19:
StatsD support it . So i made it happen in heka too. 
And this should help developer to know the how the trend really goes .

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -3,6 +3,8 @@
 Backwards Incompatibilities
 ---------------------------
 
+* StatAccumInput Input: percent_threshold param type convert to slice
+
 Bug Handling
 ------------
 
--- a/docs/source/config/inputs/stataccum.rst
+++ b/docs/source/config/inputs/stataccum.rst
@@ -23,9 +23,9 @@ Config:
     in the message fields of the generated messages. Defaults to false. *NOTE*:
     At least one of 'emit_in_payload' or 'emit_in_fields' *must* be true or it
     will be considered a configuration error and the input won't start.
-- percent_threshold (int):
+- percent_threshold (slice):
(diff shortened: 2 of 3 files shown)


i19:
The configuration should be like this

```
[StatAccumInput]
percent_threshold = [90, 95, 99]
delete_idle_stats = false
ticker_interval = 10]
```

rafrombrc:
This looks great, thanks! A couple more details need to be handled before it can be merged, however. First, we need an entry in the `Backwards Incompatibilities` section for version 0.11.0 of the CHANGES.txt changelog. Second, the documentation needs to be updated in the stataccum.rst file.

i19:
The docs has been updated, please review .

rafrombrc:
Docs looks good, thanks! Unfortunately, however, I ran the tests against your branch and one of the stat accumulator tests is now failing. Sorry for not checking that before... :P

i19:
That's a pity. 
Could you show me the testing result , or testing case, or testing method.
I didn't find the document about running the testing in the doc .

rafrombrc:
You can run `ctest` from the repo checkout directory, or `make test` from the build folder if you've already set up the environment using `. env.sh`. If the environment is set up, you can also use the go tool to just run specific package tests, e.g. `go test -v github.com/mozilla-services/heka/pipeline`.

i19:
I checked the testing case . The error is caused by the default value i set for the percentValue list .
I removed it !
And in the dev branch, when calculate the percent value, it has an logic that would judge if the timing data is bigger than 1, but if i have multi percent setting, then it would be unreasonable, and it would be conflict try to handle that situation. So i remove the judge, ether.

tclh123:
:beers:

PR #1632, opened 2015-07-23
Allow configuring a custom message type in nginx_error decoder.

jemc:
This PR makes the `nginx_error` decoder accept the `type` configuration parameter, just like the `nginx_access` decoder.  Note that the default value here is `"nginx.error"` instead of `nil` in order to keep compatibility with previous behavior of unconditionally setting the type to `"nginx.error"`.

### DIFF
--- a/sandbox/lua/decoders/nginx_error.lua
+++ b/sandbox/lua/decoders/nginx_error.lua
@@ -10,6 +10,9 @@ Config:
 - tz (string, optional, defaults to UTC)
     The conversion actually happens on the Go side since there isn't good TZ support here.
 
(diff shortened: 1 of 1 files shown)


trink:
The purpose of type in the nginx_access decoder is to differentiate between different format types.  The error log only has a single format so this is unnecessary.

jemc:
@trink - We are sending these logs through an `ESJsonEncoder` with `index = "%{Type}-%{20060102}"` to an "ElasticSearchOutput", but Elasticsearch complains when index names have the `.` character.  For us, changing the method type to `nginx-error` is necessary.

jemc:
Thanks, @rafrombrc!

PR #1636, opened 2015-07-24
add nginx stub status decoder

nathwill:
adds a lua decoder for the [nginx stub status module](http://nginx.org/en/docs/http/ngx_http_stub_status_module.html)

### DIFF
--- a/docs/source/config/decoders/index.rst
+++ b/docs/source/config/decoders/index.rst
@@ -21,6 +21,7 @@ Available Decoder Plugins
    mysql_slow_query
    nginx_access
    nginx_error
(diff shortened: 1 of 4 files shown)


nathwill:
i tagged this for 0.11 since 0.10b is already out, so i figured the merge window on 0.10 might be closed. lmk if i should do otherwise.

rafrombrc:
There's little risk here, I'm happy to have this land in 0.10, but that'll require closing this PR and opening a new one against the versions/0.10 branch. Thanks!

nathwill:
:+1:

PR #1637, opened 2015-07-24
add nginx stub status decoder

nathwill:
follow up to https://github.com/mozilla-services/heka/pull/1636


nathwill:
ah, egads, this pulled in way too much... need to rebase my commits on the 0.10 branch

PR #1638, opened 2015-07-24
add nginx stub status decoder

nathwill:
follow up to https://github.com/mozilla-services/heka/pull/1636

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,7 @@ Features
 --------
 
 * Allow multiple sandbox module directories to be specified (#1525).
+* Add Nginx stub status lua sandbox decoder
 
 Bug Fixes
 ---------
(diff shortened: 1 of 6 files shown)


rafrombrc:
This looks great, thanks! The only requests I have are an entry added to the CHANGES.txt changelog, and the addition of a couple of tests so we have a sanity check that the decoder is working. Adding tests for your decoder should be very straightforward, there are plenty of existing ones for you to use as a model. Just add yours in a new `C.Specify` call alongside the rest of them: https://github.com/mozilla-services/heka/blob/dev/sandbox/plugins/sandbox_decoders_test.go#L300.

nathwill:
ah, cool, i wasn't sure how the lua decoders were being tested, thanks for the pointer!

nathwill:
@rafrombrc alright, specs and CHANGES added, this looking better now?

PR #1641, opened 2015-07-28
Allow to set a suffix to the Debian package version

sathieu:
Example:
  CPACK_DEBIAN_PACKAGE_VERSION_SUFFIX=+deb8 make deb


sathieu:
I needed this because:
* if compiled under wheezy, it lacks the systemd services (requiring recent debhelper, which depends on dpkg-dev >= 1.17 which depends on patch >= 2.7 which depends on libc6 >=2.17 ; making the binary package depends on libc6 >=2.14)
* if compiled under jessie, it depends on libc6 (>= 2.14)
* I use reprepro which forbid duplicate names with different checksum

PR #1647, opened 2015-07-30
Postfix

sathieu:

Issue #358, opened 2013-07-31
Systemd journal inputer for Hekad


Micheletto:
Feature Request: Add a systemd journal inputer for Hekad.  See http://www.freedesktop.org/wiki/Software/systemd/ for information about systemd.

stuart-warren:
There is a pretty cool go library here for interacting with the journal https://github.com/ledbettj/systemd-journal
I used it with Logstash https://github.com/stuart-warren/logstash-input-journald

nathwill:
+1 for this request; especially as journald takes over some functions of traditional syslog, a journald input would be really useful.

ecnahc515:
@rafrombrc Alternatively you can also use `systemd-journal-gatewayd`. In any method, you'll need a way to track your position in the logstream, so maybe the cursor/offset tracking logic can be generalized somewhere?

chrissnell:
If anyone is interested in a "native" (well...with cgo) Go implementation, I'm happy to help out.  [I wrote a journald reader for another project](https://github.com/chrissnell/kube-journald-filter/blob/master/kube-journald-filter.go) with the github.com/coreos/go-system library and have been running it in production for a few months now.  I don't think it would be hard to adapt to Heka.   The one headache with of go-systemd is that newer versions of Ubuntu (and potentially others) have the journald library as part of libsystemd.  Older versions use libsystemd-journal.  So, the binaries become OS distro-specific.  Not sure if that's a problem for this project.

sathieu:
I'm not sure to understand the distro-specific part. Has the ".so" filename changed?

I currently have to build different .deb for Debian wheezy and jessie, so it won't change for me.

nathwill:
also of note here: https://www.freedesktop.org/software/systemd/man/systemd-journal-upload.html. Especially with the `--save-state` option, it seems like it ought to be a good tool to use in combination with HttpListenInput. I'll screw around with it and see if I can come up with a working pattern.

chrissnell:
@sathieu Yes, different .so files depending on which distro.  I'm not totally sure that this was a distro packaging choice--it may be systemd version-dependent.  At any rate, it's not impossible to work around and go-systemd works well.

nathwill:
lame, systemd-journal-upload's a no-go for now it seems. it performs the entire upload as one continuous streaming POST, so an HttpListenInput in heka never chunks the messages :/. sooo cloose!

so far, the ways i know of to get journal data at present include:
[...]

PR #1649, opened 2015-07-31
Add generic syslog decoder 

sathieu:
Rsyslog coupled with software-specific extract

First, it will decode according to rsyslog grammar (optional).

After, it will decode messages depending on it's programmname field.

Support for the following programs:
- PAM
- passwd (groupadd, groupdel, useradd, userdel)
- Postfix
- shadow (login)
- sshd
- su
- sudo

### DIFF
--- a/sandbox/lua/decoders/syslog.lua
+++ b/sandbox/lua/decoders/syslog.lua
@@ -1,0 +1,182 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[[
+Parses the syslog logs and extract fields of common programs.
+
+Config:
+
+- hostname_keep (boolean, defaults to false)
+    Always preserve the original 'Hostname' field set by Logstreamer's 'hostname' configuration setting.
+
+- rsyslog_template (string)
+    The 'template' configuration string from rsyslog.conf.
+    http://rsyslog-5-8-6-doc.neocities.org/rsyslog_conf_templates.html
+    If you want more flexibility, set this to nil and use MultiDecoder.
+    SyslogDecoder uses Payload and Fields[programmname] as input
+
+- tz (string, optional, defaults to UTC)
+    If your rsyslog timestamp field in the template does not carry zone offset information, you may set an offset
+    to be applied to your events here. Typically this would be used with the "Traditional" rsyslog formats.
+
+    Parsing is done by `Go <http://golang.org/pkg/time/#LoadLocation>`_, supports values of "UTC", "Local",
+    or a location name corresponding to a file in the IANA Time Zone database, e.g. "America/New_York".
+
+*Example Heka Configuration*
+
+.. code-block:: ini
+
+    [SyslogDecoder]
+    type = 'SandboxDecoder'
+    # Default 8MiB is not enough, use 16MiB
+    memory_limit = 16777216
+    filename = 'lua_decoders/syslog.lua'
+
+    [SyslogDecoder.config]
+    type = 'RSYSLOG_TraditionalFileFormat'
+    rsyslog_template = '%TIMESTAMP% %HOSTNAME% %syslogtag%%msg:::sp-if-no-1st-sp%%msg:::drop-last-lf%\n'
+    tz = 'Europe/Paris'
+
+*Example Heka Message*
+
+:Timestamp: 2014-01-10 07:04:56 -0800 PST
+:Type: RSYSLOG_TraditionalFileFormat
+:Hostname: test.example.org
+:Pid: 0
+:UUID: 8e414f01-9d7f-4a48-a5e1-ae92e5954df5
+:Logger: SyslogInput
+:Payload: 25F2E5E061: to=<george.desantis@test.example.com>, relay=none, delay=0.05, delays=0.05/0/0/0, dsn=2.0.0, status=sent (test.example.com)
+:EnvVersion:
+:Severity: 7
+:Fields:
+    | name:"programname" value:"postfix/discard"
+    | name:"postfix_queueid" value:"25F2E5E061"
+    | name:"postfix_to" value:"george.desantis@test.example.com"
+    | name:"postfix_relay" value:"none"
+    | name:"postfix_delay" value:0.05
+    | name:"postfix_delay_before_qmgr" value:0.05
+    | name:"postfix_delay_in_qmgr" value:0
+    | name:"postfix_delay_conn_setup" value:0
+    | name:"postfix_delay_transmission" value:0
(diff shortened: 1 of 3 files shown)


sathieu:
This replaces #1647.

Still WIP, I need to move part of it to lua_sandbox.

sandbox/lua/decoders/syslog.lua
```diff
@@ -0,0 +1,373 @@
...
+            msg.Hostname = fields.hostname or fields.source
+            fields.hostname = nil
+            fields.source = nil
+        end
+
+        msg.Payload = fields.msg
+        fields.msg = nil
+
+    else
+        msg = decode_message(read_message('raw'))
```
rafrombrc: Hrm, I'm a bit confused about this. `decode_message(read_message('raw'))` is for cases when we want to deserialize the protobuf encoded message bytes that are stored on the pack. Not sure why we'd want that to happen if no rsyslog template is provided... wouldn't we still be expecting to use the text data from the message payload?

rafrombrc:
Overall I like where this is going. :)

sandbox/lua/decoders/syslog.lua
sathieu: Yes, we would use the msg.Payload, but we also need msg.Fields.programname.

I've done this to preserve all metadata. I imagine in the long term a systemd-journal Input (#358) that could be used with the Syslog decoder.

rafrombrc: This still doesn't make sense. A single input / splitter combination will either store the received data as a string in the Payload field of a message struct (i.e. `pack.Message.Payload`), *or* as binary []bytes data in the MsgBytes attribute of the pack itself (i.e. `pack.MsgBytes`). This is specified by the splitter's `use_message_bytes` config setting, see http://hekad.readthedocs.org/en/v0.9.2/config/splitters/index.html.

At the point when this decoder gets the message, it will contain the raw log line in the message Payload. AFAICT, there will never be a case when using `read_message('raw')` makes sense. If the input has set a `programname` field on the message, then the field is already set, you shouldn't have to do anything in the decoder to make this happen.

sathieu: I probably have missed something. May I explain how I understand:

Given the journald "export" format (http://www.freedesktop.org/wiki/Software/systemd/export/), the splitting and decoding phase are tight together. The to-be-done systemd-journal input would behave like Docker input plugins and fill both Fields and dynamic Fields.

We should keep fields (decode_message(read_message('raw'))) and dynamic fields (fields = msg.Fields) and append more fields (fields[k] = v).

rafrombrc: I don't understand what your statement about "splitting and decoding phase are tight together" means. Let me step back a bit and explain how Heka works here in more detail:

When a Heka input gets data from the outside world, it is either in native Heka protobuf format or it isn't. If the data *is* in native Heka protobuf format, then it contains an entire Heka message. When the data is decoded, a fully populated Heka message is the result. In this case, Heka will store the raw protobuf data in the pack.MsgBytes attribute. The decoder (usually a ProtobufDecoder) will grab the data from pack.MsgBytes, will deserialize it into a Message struct, and will store that Message struct in the pack.Message attribute, overwriting any message data that may have already been there.

If the data is *not* in native Heka protobuf format, then it is clearly in some other format. Usually this format is a string of some sort. In these cases, Heka creates a mostly-empty Message struct, and stores the data in the Payload field of the Message struct. The decoder must access the message payload to get the raw input. It will then parse this input, and write the extracted data back to the same Message struct. It can populate both base fields and dynamic fields. It may overwrite the original payload, or it may not; many decoders in fact have a `payload_keep` setting that lets you configure whether or not the original payload will be preserved.

Note that in this second case, pack.MsgBytes doesn't come in to the picture at all. If the splitter did not have the `use_message_bytes` option set to true, then the decoder can *not* trust the data that is in pack.MsgBytes. There will be nothing there. All of the useful data will be in the Message struct. If the input has set dynamic message field values, those will be set on the Message struct.

When you call `read_message("Payload")` or `read_message("Fields[foo]")`, you are saying "give me the [Payload|dynamic field 'foo'] value from the `pack.Message` struct". When you call `read_message("raw")`, you are saying "give me the binary protobuf data from the pack.MsgBytes attribute".

Given all of that, here are the things that smell weird about the code that prompted this whole thread:

- All of our existing decoders expect to find their input data *either* in the pack.MsgBytes attribute *or* already in the pack.Message struct. There are no other decoders that sometimes use one, sometimes the other. There might be valid use cases for it, but they're rare, and IMO would need to be defended.
- Your conditional above is based on whether or not an rsyslog_template value was provided and, if so, whether or not it correctly compiled down to an LPEG grammar. Even if you did intend to have a decoder that was expecting its input data in two different locations, that is a horrible way to signal that behaviour.
- Folks usually use syslog for text data. I've never seen anyone use syslog as a transport for binary Heka protobuf data. It's possible, I'm sure, and someone might do it, but AFAIK it hasn't happened yet.

For all of these reasons, it seems to me that `read_message("raw")` isn't the right choice here, and that you don't quite understand what it implies. If I'm wrong about your understanding and it is your explicit intention to be handling binary protobuf data in this decoder, then I still think it's probably a bad idea.

Maybe it would help if you'd explain what behaviour you'd like to see. If the syslog template compiles, then you're parsing the input and distributing the values throughout the message you're emitting. If there's no template, or it doesn't compile, then... what?

sathieu: @rafrombrc Thanks for taking the time to explain.

I though, that read_message("raw") would contain all the fields, encoded, event when not using MsgBytes.

This encoder is not intended to be used with MsgBytes, the idea is to get all the fields from the original message (i.e Payload, but also Pid, ... and dynamic fields). This could be usefull if the input fills them (i.e. the hypothetic systemd-journal input).

Additionally, the decoder should fail if the rsyslog grammar does not match.

Will push an updated version.

sathieu:
@rafrombrc What do you think now?

sandbox/lua/decoders/syslog.lua
rafrombrc: `read_message("raw")` *always* access pack.MsgBytes. That's exactly what it does, no more, no less.

What's probably confusing you is that Heka will always create a protobuf encoding of every message and store that in pack.MsgBytes before delivering the message to the router. This means that (unless there are bugs) you can always trust `read_message("raw")` to give you the correct data in any plugin that is *after* the router, i.e. any filter, encoder, or output plugin. Inputs and decoders cannot make this assumption, however.

All message fields that are set by an input are available in the decoder, but you have to use `read_message(<fieldname>)` to get them one at a time, you can't use `read_message("raw")` except in very specific cases.

sathieu: I'm still a little bit confused. What do you think about the newer version of the patch?

sathieu:
Replaced by #1760 on the 0.10 branch.

PR #1652, opened 2015-07-31
SIGUSR2 to serialize sandboxes when wedged

rafrombrc:


pipeline/pipeline_runner.go
```diff
@@ -386,3 +399,123 @@ func Run(config *PipelineConfig) {
...
+
+	// If we got this far the router is presumably wedged and we'll perform the
+	// operation.  First we close the abort channel to allow any hung
+	// inject_message calls to return, then we go through and stop and destroy
+	// all of the sandboxes. Unfortunately the data structures we use to track
+	// the sandbox plugins of each type are just different enough that it's
+	// hard to write generic code, so we loop through each plugin type
+	// separately.
+	close(config.Globals.abortChan)
+	time.Sleep(1 * time.Second)
```
trink: Arbitrary 'waits' for stuff to finish in a multi threaded environment will eventually hit a race condition.  If we cannot just start destroying we probably need a better solution.

sandbox/lua/lua_sandbox_interface.c
```diff
@@ -100,13 +104,18 @@ int timer_event(lua_sandbox* lsb, long long ns)
...
         char err[LSB_ERROR_SIZE];
         size_t len = snprintf(err, LSB_ERROR_SIZE, "%s() %s", func_name,
-                              lua_tostring(lua, -1));
+                              errmsg);
         if (len >= LSB_ERROR_SIZE) {
           err[LSB_ERROR_SIZE - 1] = 0;
         }
-        lsb_terminate(lsb, err);
+        // Don't terminate if we're aborting so we preserve data on exit.
+        const char* tail = errmsg + strlen(errmsg) - 7; // Last 7 chars of err msg.
```
trink: Out of bounds if the error message is less than seven characters long (you can use lua_tolstring to grab the length when retrieving errmsg or just grab it off the end of err (which is always longer than 7)

sandbox/plugins/sandbox_decoder.go
```diff
@@ -176,6 +176,9 @@ func (s *SandboxDecoder) SetDecoderRunner(dr pipeline.DecoderRunner) {
 	s.sb.InjectMessage(func(payload, payload_type, payload_name string) int {
 		if s.pack == nil {
 			s.pack = dr.NewPack()
+			if s.pack == nil {
+				return 1 // We're aborting, exit out.
```
trink: Abort is rc = 5

sandbox/plugins/sandbox_input.go
```diff
@@ -115,7 +122,10 @@ func (s *SandboxInput) Run(ir pipeline.InputRunner, h pipeline.PluginHelper) (er
 			ct, _ := time.ParseInLocation(layout, t.Format(layout), s.tz)
 			pack.Message.SetTimestamp(ct.UnixNano())
 		}
-		ir.Inject(pack)
+		if err := ir.Inject(pack); err != nil {
+			pack.Recycle(nil)
+			return 1
```
trink: This and line 116 above should return abort (5)

Issue #1651, opened 2015-07-31
Provide generic JSON decoder


rafrombrc:
Be nice to provide a decoder that parses arbitrary json, flattens it down so there aren't any nested objects, and writes the data out into Heka message fields.

PR #1653, opened 2015-08-01
add generic json lua decoder

nathwill:
addresses https://github.com/mozilla-services/heka/issues/1651. i wasn't sure about whether it would be preferable or not to add the mapping to heka message fields, so went ahead and added some of the more directly applicable ones (type, severity...). if we want this to be more generic, i can pull those out.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -13,6 +13,9 @@ Bug Handling
 
 * Add sample rate support for gauge type in StatsdInput.
 
+* Fixed some cases where SandboxDecoder wasn't correctly setting
+  `pack.TrustMsgBytes` to false when needed.
+
 * Fixed broken queue rolling and added test.
 
 * Force ElasticSearch index name to lower case, as required by ElasticSearch.
--- a/sandbox/lua/lua_sandbox.go.in
+++ b/sandbox/lua/lua_sandbox.go.in
@@ -358,6 +358,7 @@ func go_lua_write_message_string(ptr unsafe.Pointer, c, v, rep *C.char,
 		lsb.globals.LogMessage("go_lua_write_message_string", "No sandbox pack.")
 		return 1
 	}
+	lsb.pack.TrustMsgBytes = false
 	if !lsb.messageCopied && lsb.sbConfig.PluginType == "encoder" {
 		lsb.pack.Message = message.CopyMessage(lsb.pack.Message)
 		lsb.messageCopied = true
@@ -458,6 +459,7 @@ func go_lua_write_message_double(ptr unsafe.Pointer, c *C.char, v C.double, rep 
 		lsb.globals.LogMessage("go_lua_write_message_double", "No sandbox pack.")
 		return 1
 	}
+	lsb.pack.TrustMsgBytes = false
 	if !lsb.messageCopied && lsb.sbConfig.PluginType == "encoder" {
 		lsb.pack.Message = message.CopyMessage(lsb.pack.Message)
 		lsb.messageCopied = true
@@ -500,6 +502,7 @@ func go_lua_write_message_bool(ptr unsafe.Pointer, c *C.char, v bool, rep *C.cha
 		lsb.globals.LogMessage("go_lua_write_message_bool", "No sandbox pack.")
 		return 1
 	}
+	lsb.pack.TrustMsgBytes = false
 	if !lsb.messageCopied && lsb.sbConfig.PluginType == "encoder" {
 		lsb.pack.Message = message.CopyMessage(lsb.pack.Message)
 		lsb.messageCopied = true
@@ -525,6 +528,7 @@ func go_lua_delete_message_field(ptr unsafe.Pointer, c *C.char, fi, ai int, has_
 		lsb.globals.LogMessage("go_lua_delete_message_field", "No sandbox pack.")
 		return 1
 	}
+	lsb.pack.TrustMsgBytes = false
 	if !lsb.messageCopied && lsb.sbConfig.PluginType == "encoder" {
 		lsb.pack.Message = message.CopyMessage(lsb.pack.Message)
 		lsb.messageCopied = true
--- a/sandbox/plugins/sandbox_decoder.go
+++ b/sandbox/plugins/sandbox_decoder.go
@@ -209,8 +209,10 @@ func (s *SandboxDecoder) SetDecoderRunner(dr pipeline.DecoderRunner) {
 				t = t.In(time.UTC)
 				ct, _ := time.ParseInLocation(layout, t.Format(layout), s.tz)
 				s.pack.Message.SetTimestamp(ct.UnixNano())
+				s.pack.TrustMsgBytes = false
 			}
 		} else {
+			s.pack.TrustMsgBytes = false
 			s.pack.Message.SetPayload(payload)
 			ptype, _ := message.NewField("payload_type", payload_type, "file-extension")
 			s.pack.Message.AddField(ptype)


sandbox/lua/decoders/json.lua
```diff
@@ -0,0 +1,106 @@
...
+    local ok, json = pcall(cjson.decode, read_message("Payload"))
+    if not ok then
+        return -1
+    end
+
+    if payload_keep then
+        msg.Payload = read_message("Payload")
+    end
+
+    if json["type"] then
```
rafrombrc: Tiny little style nit: when looking up values with a static table key, the usual Lua idiom is dot notation, i.e. `if json.type then` instead of `if json["type"] then`.

sandbox/lua/decoders/json.lua
```diff
@@ -0,0 +1,106 @@
...
+        msg.Type = json["type"] ; json["type"] = nil
+    end
+
+    if json["severity"] then
+        msg.Severity = json["severity"] ; json["severity"] = nil
+    end
+
+    if json["hostname"] then
+        msg.Hostname = json["hostname"] ; json["hostname"] = nil
+    end
```
rafrombrc: Rather than hard-coding a check for 'type', 'severity', and 'hostname', I'd like to see a data structure that contains all of the reserved message fields. From the [message schema docs](http://hekad.readthedocs.org/en/v0.10.0b0/message/index.html), the fields I think we want to check for are uuid, timestamp, type, logger, severity, payload, env_version, pid, and hostname. I imagine iterating through the top level keys of the deserialized JSON, lowercasing each value, and checking to see if it's in the set of default fields. If so, try to convert to the right type (timestamp in particular will need to be handled carefully) and then set the value on the msg, removing it from the source table as you're already doing.

sandbox/lua/decoders/json.lua
```diff
@@ -0,0 +1,106 @@
...
+
+    local flat = {}
+    util.table_to_fields(json, flat, nil)
```
rafrombrc: It doesn't look like `table_to_fields` is accounting for keys that already contain a "." in the name, which would then cause problems if someone tried to reverse the operation. Probably a good idea to tweak `table_to_fields` to optionally replace any "."s with "_"s in the key names, and then adding some tests with nested objects and dotted key names to verify that things come out right.

rafrombrc:
This looks great, thanks for sharing! I made some notes that unfortunately add a bit of work to this. Hopefully you're up to tackling it. If not, please let us know. :)

sandbox/lua/decoders/json.lua
trink: Also a type check should be used instead of an existence check. i.e., `type(json.hostname) == "string"`

sandbox/lua/decoders/json.lua
trink: if the point is to make it reversible then escaping and not substitution would be needed.

rafrombrc: Good point. I was intending to be able to at least get the nesting levels correct, not necessarily have the reversed names match exactly, but escaping is probably a better idea.

sandbox/lua/decoders/json.lua
nathwill: :+1: will do

sandbox/lua/decoders/json.lua
nathwill: so, sounds like we want to escape `.`, and not worry about reversability since there's already an option to keep the original payload intact?

sandbox/lua/decoders/json.lua
nathwill: ah, didn't know that, thanks.

nathwill:
@rafrombrc @trink ok, i expanded the mapping of the enumerated fields from the json keys to the relavent heka fields (type appropriate). the timestamp is set up to handle things like the gelf decoder: to translate from a timestamp in seconds from unix epoch to ms. thoughts?

sandbox/lua/decoders/json.lua
```diff
@@ -0,0 +1,116 @@
...
+    -- keep payload, or not
+    if payload_keep then
+        msg.Payload = read_message("Payload")
+    end
+
+    -- map to heka native message fields
+    local field_type_map = {
+        string = { "Uuid", "Type", "Logger", "Hostname" },
+        number = { "Severity", "EnvVersion", "Pid" }
+    }
```
rafrombrc: This is a static value, you want to define it outside of `process_message` or else it will need to be allocated (and subsequently GCed) for every message.

sandbox/lua/decoders/json.lua
```diff
@@ -0,0 +1,116 @@
...
+
+    for k, v in pairs(field_type_map) do
+        for i, f in ipairs(v) do
+            local low_f = f:lower()
+
+            if type(json[low_f]) == k then
+                msg[f] = json[low_f] ; json[low_f] = nil
+            end
+        end
+    end
```
rafrombrc: Not quite sure what's going on in here. Seems like we're lower-casing the values in the `field_type_map` table, but we hard-coded those values, we could have just made them lower case in the first place.

What we really want to do is to iterate through the top level values in the `json` table. We lower case every key, check to see if the lower-case value matches one of the lower-case message field names, with the right type. If so, that data gets written to that message field instead of as a dynamic field.

This will clearly cost some cycles, especially when the incoming JSON structure is large, so it might be nice to be able to turn it off with a config setting.

sandbox/lua/decoders/json.lua
nathwill: ah, yes, good call

sandbox/lua/decoders/json.lua
nathwill: they're uppercased since that's how they're represented in heka; if we don't store them uppercase, we have to capitalize them when assigning to the message table, so nothing's saved either way. i favored lower-casing them when checking the json table since that's how data's typically represented in json (lowercase).

i don't see the point of iterating over every key in the json table (like you said, this could be really slow) to check for a fixed set of known heka fields. am i missing something?

nathwill: OK, i think i see where this is confusing.

> If so, that data gets written to that message field instead of as a dynamic field.

I think that's what's already happening; see the `pack.Message.GetSeverity()` and `pack.Message.GetHostname()` specs I added to confirm.

nathwill: Oh, oh. For some reason this finally clicked in the middle of a baseball game... Correct me if I'm wrong, but I think the difference here is that the way you're proposing is case insensitive wrt the json keys, whereas the current implementation relies on the keys being lowercase (the most common in the real world in my experience).

I like the idea of being case-insensitive, but the current implementation is a lot faster than any case-insensitive version would have to be.

As I see it, there's three possible options here for the json keys:

- all lower case: covered by the current implementation in the for loop
- capitalized: covered by the current implementation in the direct assignment at the end
- camel case/other: covered by the proposed change, but slow as it's require iterating over all of the json top-level keys

So, what if we added a "downcase_fields" boolean config option (default false), and, if true,  do the downcasing *before* the field_type_map mapping, so folks could opt into the slower implementation by normalizing their keys, but we otherwise still have a decent chance of doing an efficient mapping for the first two cases?

nathwill: ok, just for kicks i added a downcase_fields implementation so we could see what that'd look like. i look forward to your comments!

nathwill:
ok, rewrote this to reflect your recommendations, and i agree it's definitely a cleaner way to do this. holler if there's anything else you'd like to see changed :)

sandbox/lua/decoders/json.lua
rafrombrc: Sorry, been on vacation (and blissfully off-line) for the last week... and am currently in for 3 days before being out for another short jaunt. Yes, I think the config option is a fine idea, once I get caught up on the email backlog I'll circle back around and review the latest code.

nathwill: no worries, there's no rush at all. enjoy your vacation!

rafrombrc:
First, sorry for the delay, I'm on a 3-day work week wedged in between two chunks of time away from computer, so things have been (and will continue to be) moving slowly in Heka-land.

Been thinking about this more and it occurs to me that the best choice is to add a `field_map` setting that lets the user specify which JSON fields should be mapped to which message fields. The user will provide a mapping where the keys are the (correctly capitalized) Heka message field names, and the values are the names of the JSON keys that should populate the message fields.

The `map_fields` option could stay, so if `map_fields` was false, `field_map` would have no effect. If `map_fiels` is true and no `field_map` is specified, the default `field_map` would be used, which might look like this:

```json
{"Payload": "payload",
 "Uuid" : "uuid",
 "Type": "type",
 "Logger": "logger",
 "Hostname": "hostname",
 "Severity": "severity",
 "EnvVersion": "env_version",
 "Pid": "pid",
 "Timestamp": "timestamp"}
```

Could even support multiple values on the right side for inconsistent JSON, so this would be valid:

```json
{"Payload": ["Payload", "payload"],
 "Uuid" : ["Uuid", "uuid"],
 "Type": ["Type", "type"],
 "Logger": ["Logger", "logger"],
 "Hostname": ["Hostname", "hostname"],
 "Severity": ["Severity", "severity]",
 "EnvVersion": ["EnvVersion", "env_version"],
 "Pid": ["Pid", "pid"],
 "Timestamp": ["Timestamp", "timestamp"]}
```

This gets rid of the iteration through the JSON fields and the need to lowercase the field names. The only awkward part of this is that the sandbox config doesn't support nested TOML sections, settings can only be boolean, number, or string values. So `field_map` will have to be a JSON string that the decoder parses to get the mapping.

Hopefully this all makes sense.

nathwill:
hey, rob. no worries, hope you're enjoying your time off. 

i think i understand, and i like the idea of a user-provided mapping, though i'm not super keen about supporting multiple fields mapping to heka fields; it's extra complexity that seems unlikely to be encountered in the real world from a single data source (and if it's multiple sources, just set up another input->decoder in the heka config), and... what do we do if both fields exist? staying simple and clear seems preferable. i'm also still not clear on how we can offer both keep_payload and (p|P)ayload field mapping, so for now i'll just write in keep_payload support to keep closer parity with the other decoders.

in any case, i'll start working on rewriting this to support user-provided mappings like you described.

nathwill:
@rafrombrc ok, i think this matches what you described, excepting my stated omissions. how's this looking?

PR #1658, opened 2015-08-04
Update payload encoder docs to reflect use of strftime syntax to

rafrombrc:
specify timestamp format.


docs/source/config/encoders/payload.rst
```diff
@@ -41,4 +39,4 @@ Example
 	[PayloadEncoder]
 	append_newlines = false
 	prefix_ts = true
-	ts_format = "2006/01/02 3:04:05PM MST"
+	ts_format = "%Y/%m/%d %l:%M:%S%p %Z"
```
trink: %l ??

PR #1661, opened 2015-08-04
Protobuf encode the report messages before injecting them

### LINKED ISSUE #1659: heka.memstat messages may not generate Protobuf representation
In 0.10.0b0, `read_message('raw')` returns nil when heka.memstat messages are given to a sandbox output.

rafrombrc:
(fixes #1659).


trink:
+1 after fixing up the conflict

PR #1663, opened 2015-08-05
Fix hostname bug in the Graylog decoder

carlanton:
I found a bug in the Graylog decoder! This should work better :-)


rafrombrc:
I've cherry-picked your commit and merged it into the versions/0.10 branch so it will go out in the next release. The change will be merged back to dev soon, closing this PR. Thanks!

PR #1665, opened 2015-08-06
make stats_input  gauge type support rate setting

i19:


rafrombrc:
Thanks! I've cherry-picked your commit and merged it into the versions/0.10 branch so it will go out in the next release.

PR #1672, opened 2015-08-10
Support sd_notify (from systemd) for reliable service status

sathieu:
See http://www.freedesktop.org/software/systemd/man/sd_notify.html

### DIFF
--- a/packaging/debian/heka.service.in
+++ b/packaging/debian/heka.service.in
@@ -4,6 +4,7 @@ After=network.target auditd.service
 ConditionPathExists=!/etc/heka/hekad_not_to_be_run
 
 [Service]
+Type=notify
 EnvironmentFile=-/etc/default/heka
 User=heka
 Group=heka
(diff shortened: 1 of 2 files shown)


sathieu:
This is an interim patch.

As better integration would include
* systemd-journal input (#358),
* sd_notify (this one)
* socket-activation (LISTEN_PID/LISTEN_FDS)
* journal output

Maybe using https://github.com/coreos/go-systemd would be a better path than rewriting everything. What do you think? (even if journal reading is not done yet: https://github.com/coreos/go-systemd/issues/70)

trink:
See: https://github.com/daurnimator/lua-systemd

sathieu:
Thanks @trink, but how to do socket activation of an UdpInput (written in go) from a lua library?

trink:
I am not sure what your use case is or even why you would need/want to split it up between Go/Lua

sathieu:
I need to:
*  have a journal input,
* socket-ativate port 514 (udp and tcp), to loose less syslog messages
* Have Type=notify, to correctly handle hekad availibility
* (optionnaly) have watchdog

I don't care if this is done in lua or Go, but point 2 and 3 probably need to be in Go as they are in heka core.

sathieu:
Closing PR as Heka is a dead projet...

PR #1677, opened 2015-08-14
Force ES index name to lower case

sathieu:
See https://www.elastic.co/guide/en/elasticsearch/guide/current/_document_metadata.html#_index:
This name must be lowercase, cannot begin with an underscore, and cannot contain commas.


rafrombrc:
Added changelog text and merged this into the [versions/0.10 branch](https://github.com/mozilla-services/heka/commit/743152642c0c81bce4209f371dd8c671a4274ff6) so it will make it into the next release.

PR #1684, opened 2015-08-19
Missed some places where SandboxDecoder should ask for a re-encoding

rafrombrc:


sandbox/plugins/sandbox_decoder.go
```diff
@@ -324,6 +326,7 @@ func (s *SandboxDecoder) Decode(pack *pipeline.PipelinePack) (packs []*pipeline.
 	if retval == 0 && s.pack != nil {
 		// InjectMessage was never called, we're passing the original message
 		// through.
+		pack.TrustMsgBytes = false // The message was presumably mutated.
```
trink: The flag could be set in go_lua_write_message* to avoid re-encoding a message being passed through as-is.

rafrombrc: Good point. Done.

Issue #1218, opened 2014-12-10
Lua Sandbox Timestamp Parsing Documentation


hmrm:
In the documentation for "write_message" in the Lua sandbox (https://hekad.readthedocs.org/en/latest/sandbox/index.html#lua-sandbox), it mentions that "Timestamp (accepts Unix ns-since-epoch number or a handful of parseable string representations.)" It would be helpful to have those formats specified or a link to where one could find them.

PR #1687, opened 2015-08-20
add generic json sandbox lua decoder (fixes #1651)

### LINKED ISSUE #1651: Provide generic JSON decoder

nathwill:
submitting new PR to clean up the git history. see #1653 for prior discussion. addresses #1651


ChristianKniep:
+1

nathwill:
@ChristianKniep there's nothing special about this decoder's use of the cjson module. the cjson module is used by a number of the sandbox plugins (e.g. the gelf decoder), and the module is included as part of the normal heka distribution when you install from a release package or `source build.sh`. If you want to use this plugin before it's merged, just dropping it into `/usr/share/heka/lua_decoders` on a machine with a pre-installed heka package works for me.

nathwill:
ah, ok, it comes via [this](https://github.com/mozilla-services/heka/blob/dev/cmake/externals.cmake#L21) and [this](https://github.com/mozilla-services/lua_sandbox/blob/dev/cmake/externals.cmake#L86)

rafrombrc:
@ChristianKniep You probably need to run `make install` and then set your `share_dir` in the global `[hekad]` config section to be `/path/to/heka/build/heka/share/heka`.

sandbox/lua/decoders/json.lua
```diff
@@ -0,0 +1,170 @@
...
+    Type     = msg_type,
+    Payload  = nil,
+    Fields   = nil,
+    Severity = nil
+}
+
+function process_message()
+    local ok, json = pcall(cjson.decode, read_message("Payload"))
+    if not ok then
+        return -1
```
rafrombrc: You can return a second argument that will be logged as an error message, be nice to add those to the error returns.

sandbox/lua/decoders/json.lua
```diff
@@ -0,0 +1,170 @@
...
+    expects field value to be numeric.
+
+- Pid (string, optional, default nil)
+    string specifying json field to map to message Pid,
+    expects field value to be numeric
+
+- Timestamp (string, optional, default nil)
+    string specifying json field to map to message Timestamp,
+    expects field value to be unix epoch timestamp in seconds.
+
```
rafrombrc: Might be nice to support `Payload`, which would override `payload_keep` like `Type` overrides `type`.

sandbox/lua/decoders/json.lua
```diff
@@ -0,0 +1,170 @@
...
+            if type(json[f]) == field_type_map[F] then
+                if F == "Timestamp" then
+                    msg[F] = json[f] * 1e9
+                else
+                    msg[F] = json[f]
+                end
+                json[f] = nil
+            end
+        end
+    end
```
rafrombrc: I think we're leaking values through here. If one message has a value that comes from the field map but the next message is missing that key in the JSON, then unless I'm missing something the `msg` value will still hold the previous message's value. We should add all of the possible fields in use to the `msg` initialization, and then be sure to explicitly nil out or set defaults for all static message fields that we don't get from the input JSON.

sandbox/lua/decoders/json.lua
```diff
@@ -0,0 +1,170 @@
...
+
+    -- flatten and assign remaining fields to heka fields
+    local flat = {}
+    util.table_to_fields(json, flat, nil)
+    msg.Fields = flat
+
+    if not pcall(inject_message, msg) then return -1 end
```
rafrombrc: Another place we can return an error message.

ChristianKniep: Thanks Rob,

I will give it a try and report back.
Thanks a bunch for this extended answer, I appreciate it.

Cheers from Berlin
Christian

sandbox/lua/decoders/json.lua
nathwill: > I think we're leaking values through here.

My bad; I guess I misunderstood what happens between decodes; I assumed the initialization before the process message function was run for every decode, but it sounds like that's not the case.

>We should add all of the possible fields in use to the msg initialization
Gotcha.

> and then be sure to explicitly nil out or set defaults for all static message fields that we don't get from the input JSON.

If I'm correctly interpreting the suggestion here, wouldn't it be easier to just set up the initial values for the msg table at the top of the process_message function at this point?

sandbox/lua/decoders/json.lua
nathwill: okey doke. added.

sandbox/lua/decoders/json.lua
rafrombrc: If we define `msg` entirely within process_message then it will be allocated every time process_message is called and it will go out of scope every time process_message exits, leading to a lot of unnecessary GC churn. When we define it at the module scope, initializing all of the keys we're going to want with a `nil` value, then the memory space for the Lua table will be created once and reused each time. It does require us to clear the values out each time to prevent the values from leaking through from one message to the next, but the performance improvement is worth it.

nathwill: ok, very interesting, i'd have thought the extra cycles looping over the keys a second time would outweigh the GC/memalloc penalty, but it sounds like i'm wrong there. thanks for the explanation!

nathwill: alright, i think i found a reasonably efficient way to do this, but if you know a better way, i'm all ears :)

rafrombrc: I think we can do this without the additional loop, can't we? For static fields, the only values that will ever be set are ones that are in the field map. If it's not in the field map, it's initialized as `nil` and never changes, so we don't have to reset it each time. If it is in the field map, then we can check if the value is in the JSON. If so, we set it to the extracted value; if not, or if the type is wrong, we set it to `nil`.

nathwill: if we get rid of the `type` config option, yeah i think that'd work fine.

rafrombrc: Do we have to get rid of `type`? If `type` is set but `Type` isn't used, then msg.Type will be initialized once at the beginning and never change, just as with `nil` for the other values. If `Type` is used, then the msg.Type will be overwritten with either the JSON value or `nil`, which matches the documented behaviour.

nathwill: hmmm, yeah, i suppose that's true, i was thinking if `type` and `Type` config options were set, we'd want to persist `type`, but i guess the docs would imply that `Type` could end up as nil if set and not present.

rafrombrc: If you really want to default to `type` when `Type` fails to return a value then it could be special cased in the loop, would only cost one string compare per field that is set. I don't feel strongly about what the behaviour should be here.

sandbox/lua/decoders/json.lua
```diff
@@ -0,0 +1,187 @@
...
+    -- keep payload, or not
+    if payload_keep then
+        msg.Payload = read_message("Payload")
+    end
+
+    -- map fields
+    if map_fields then
+        for F, f in pairs(field_map) do
+            if type(json[f]) == field_type_map[F] then
+                if F == "Timestamp" then
```
nathwill: i actually had a question about this; i copied what the gelf decoder was doing here, but i'm not super enthusiastic about either locking the user into a specific format (epoch), or trying to do a bunch of crazy date parsing. i was searching through the open issues for somewhere i might be able to help and maybe get a little more familiar with the internals, and ran into #1218, which seems to imply that there may be some support for a specific set of timestamp formats without having to parse them in every decoder. if so, maybe we could just document "accepts mapped timestamps in these supported formats"; skip the special case, and give users more options...

so, do lua decoders support passing timestamps in any of [these](https://github.com/mozilla-services/heka/blob/609bd9bc36de249d823e4f95f025415ff46249dd/message/date_helpers.go#L26-L42) formats? if so, i'd be happy to add that list to the docs referenced in #1218, get this decoder updated to support the full range of heka-supported timestamp formats.

sandbox/lua/decoders/json.lua
nathwill: meh. i'll leave it; simpler and less potential to be surprising by being too clever that way.

sandbox/lua/decoders/json.lua
rafrombrc: Yes, sorry the docs could use quite some improvement here. We provide a date_time module in the Lua sandbox that allows you to parse date strings if you provide the correct strftime format representation string. Calling [date_time.build_strftime_grammar](https://github.com/mozilla-services/lua_sandbox/blob/v0.7.0/modules/date_time.lua#L217) returns an LPEG grammar that does the parsing for you. It returns a Lua table that can be turned into ns-since-epoch using [date_time.time_to_ns](https://github.com/mozilla-services/lua_sandbox/blob/v0.7.0/modules/date_time.lua#L13). You can add an optional timestamp_format config setting where users can specify the strftime format string. Lack of a timestamp_format setting would presume the default timestamp format of ns-since-epoch, i.e. exactly what we have now.

nathwill: rad, thanks. i'll take a look later this afternoon and see if i can sort that out.

PR #1708, opened 2015-09-04
Exit for-loop when stopChan closes

### LINKED ISSUE #1705: SandboxInput with a "ticker_interval" will not shutdown
When using a SandboxInput that has a `ticker_interval` and attempting to shutdown Heka, the plugin will get stuck in an infinite tight loop where the Lua script is re-executed as soon as it returns.

The issue can be reproduced with the following TOML configuration and associated Lua script: https://gist.github.com/TheDude05/f2feccfbe282132e086a. If you start hekad with this configuration and send it a SIGINT, you'll see the Lua script infinitely write to stderr until you forcibly SIGKILL the process.

I brought this up in IRC and @trink pointed out this problematic line: https://github.com/mozilla-services/heka/blob/dev/sandbox/plugins/sandbox_input.go#L151. I'm creating this issue per that discussion.

TheDude05:
Fixes #1705


rafrombrc:
This looks great, just a few points:
- would love to have a CHANGES.txt update
- would love to have it opened against `versions/0.10` branch so it will land in the next release
- you could actually avoid the label if you changed `for true` to `for ok`, I don't feel strongly enough to force it, though, up to you.

TheDude05:
That should be fine. I'll open a new pull request as I don't think Github will let me change the branch to merge into once the pull request has been created.

For clarification on point 3, did you mean do something like this?

```
ok := true
for ok {
    // Do some things

        select { // block until stop or poll interval
        case _, ok = <-s.stopChan:
        case <-ticker:
        }
```

rafrombrc:
Yes, exactly.

TheDude05:
Created #1709 with the suggested changes.

Issue #1066, opened 2014-09-08
LogstreamerInput behavior change with non-existent directory


whd:
This configuration will run without error in 0.6.0-1 (where /nonexistent is indeed nonexistent):

```
[Test]
type = "LogstreamerInput"
log_directory = "/nonexistent"
file_match = 'test\.log'
decoder = "ProtobufDecoder"
```

If the directory is later created and the log populated logstreamer picks up the new stream. However, in 0.7.1-1, the following fatal error is observed:

```
2014/09/08 13:36:23 Pre-loading: [Test]
2014/09/08 13:36:23 Pre-loading: [ProtobufDecoder]
2014/09/08 13:36:23 Pre-loading: [ProtobufEncoder]
2014/09/08 13:36:23 Loading: [ProtobufDecoder]
2014/09/08 13:36:23 Loading: [ProtobufEncoder]
2014/09/08 13:36:23 Loading: [Test]
2014/09/08 13:36:23 Initialization failed for 'Test': lstat /nonexistent: no such file or directory
2014/09/08 13:36:23 Error reading config: 1 errors loading plugins
```

This new behavior actually helped us find a misconfiguration in our staging system, but it may not be a desirable default behavior. If it is, it should be labeled as a breaking change. I would hazard to guess that #741  is where the new behavior was introduced.

PR #1717, opened 2015-09-09
Document explicitly that logstreamer's log_directory must exist

ioc32:
Way back logstreamer's default behavior changed (#1066) and this was
addressed by Rob in #1132 by amending CHANGES.txt.

This diff makes the behavior clearer for new-comers by making the note
in the main documentation.


sanga:
+1 this bit me for a moment too

PR #1723, opened 2015-09-11
Fix Influx parsing and add batching filter

### LINKED ISSUE #1718: batch output or encoder for influxdb
batch output will gain great performance 
is it  possible to write sandboxFilter filter to generate the multi line payload then using the payload encoder  with the httpout ?

rafrombrc:
Also adds tests for newer influx parsing. Fixes #1718.

### DIFF
--- a/sandbox/lua/filters/influx_batch.lua
+++ b/sandbox/lua/filters/influx_batch.lua
@@ -173,7 +173,7 @@ function process_message()
     -- Inject a new message with the payload populated with the newline
     -- delimited data points, and append a newline at the end for the last line
     any = true
-    add_to_payload(table.concat(api_message, "\n").."\n")
+    add_to_payload(table.concat(api_message, "\n"), "\n")
 
     if flush_count > 0 then
         count_since_flush = count_since_flush + 1
--- a/sandbox/lua/filters/influx_batch.lua
+++ b/sandbox/lua/filters/influx_batch.lua
@@ -1,0 +1,190 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[=[
+
+Converts Heka message contents to InfluxDB v0.9.0 API format and periodically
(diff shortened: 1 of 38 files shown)


sandbox/lua/filters/influx_batch.lua
```diff
@@ -0,0 +1,190 @@
...
+    end
+    count_since_flush = 0
+end
+
+function process_message()
+    local api_message = ts_line_protocol.influxdb_line_msg(config)
+    -- Inject a new message with the payload populated with the newline
+    -- delimited data points, and append a newline at the end for the last line
+    any = true
+    add_to_payload(table.concat(api_message, "\n").."\n")
```
trink: comma instead of .. will prevent as second string from being created

sandbox/lua/modules/ts_line_protocol.lua
```diff
@@ -148,23 +147,40 @@ local function points_tags_tables(config)
...
-                and (config["tag_fields_all"]
-                or (config["used_tag_fields"] and used_tag_fields[field])) then
-                    table.insert(tags, influxdb_kv_fmt(field).."="..tostring(influxdb_kv_fmt(value)))
+            field_out_name = field
+
+            field_idx = seen_fields[field] or -1
+            field_idx = field_idx + 1
+            seen_fields[field] = field_idx
+            if field_idx > 0 then
+                field_out_name = field_out_name .. "_fidx_" .. tostring(field_idx)
```
trink: Prefer string.format for anything over a single concatenation (your call, I would have to profile it to see if it is worth it)
```lua
field_out_name = string.format("%s_fidx_%d", field_out_name, field_idx)
```

rafrombrc: Yeah, I did some profiling a while back and found that the balance seemed to tilt towards string.format performing better than concatenation when the string count got to about 5. Not sure if real world usage would match that. I'm happy to settle on string.format for anything over a single concatenation.

peternga:
Is it possible to include a configuration parameter to specify tag fields to skip? I've made a change that uses skip_fields to affect tag_fields but I think a cleaner way to approach this is to simply have another parameter that only affects tag_fields.

PR #1724, opened 2015-09-12
Account for the case where the file that matches our lookup id is

rafrombrc:
created after we look for it directly but before we get the full list
of files from the queue folder.


pipeline/queue_buffer.go
```diff
@@ -329,7 +331,7 @@ func (br *BufferReader) getFileFromId(id uint) (file *os.File, foundId uint,
 
 	// Increment until we find an id greater than what was requested.
```
trink: comment no longer matches the code

PR #1703, opened 2015-09-02
Update to Sarama 1.5.0, add `git_clone_to_path` to build, use it

rafrombrc:
to clone Sarama into its Shopify package path.

PR #1727, opened 2015-09-14
Upgrade to Sarama 1.6

andremedeiros:
This PR updates Sarama to use the Shopify mainstream and to its latest version.


rafrombrc:
Unfortunately, there was already an open PR (#1703) with an upgrade to Sarama 1.5, along with updating all of the tests, so you've duplicated work that was already done. :P Sorry about that!

PR #1735, opened 2015-09-22
Read the whole http response.

frison:
Without requiring the full response is read prior to returning, the
closing of the response body might not work, resulting in many sockets in
the TIME_WAIT state.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -11,6 +11,11 @@ Bug Handling
 * Updated Sarama dependency from pre-1.0 release fork to fork (with only test
   code changes) of Sarama 1.5.0 release.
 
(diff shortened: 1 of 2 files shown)


frison:
I haven't the faintest clue why Travis failed on this guy. All I know is that it fixed issues we were seeing on our windows machines. I'll dig into it soon, any guidance would be appreciated.

rafrombrc:
We have ongoing issues with intermittent Travis failures, mostly related to (test only) timing issues that are hard to debug because they only happen in Travis and not locally. I've resolved some of them recently, but there are still more. I re-ran the tests and they passed the next time through, so you're fine there.

This change looks great, thanks for your contribution! The only extra piece needed is an entry in the CHANGES.txt changelog in the "bug fixes" section for v0.11.

PR #1741, opened 2015-09-24
Added `print_decode_failures` config value to decoders.

sathieu:


sathieu:
Not sure what to do about sendFailure

(ref: https://github.com/mozilla-services/heka/blob/versions/0.10/pipeline/plugin_runners.go#L443)

sathieu:
NB: I need this because, under systemd, all stdout/stderr is logged (and forwarded to syslog, which means I have the info twice, the printed form being subobtimal)

sathieu:
WIll do it again on v0.10 branch

sathieu:
replaced by #1742

PR #1742, opened 2015-09-24
Added `print_decode_failures` config value to decoders.

sathieu:


sathieu:
NB: I need this because, under systemd, all stdout/stderr is logged (and forwarded to syslog, which means I have the info twice, the printed form being subobtimal)

rafrombrc:
I don't love love love this. Ultimately, I feel like all plugin types should grow some more advanced logging control, specified by a `[PluginName.logging]` config section, and enforced in the plugin runners' LogMessage and LogError methods. That would give the ability to turn off writing errors to StdErr for all plugins, not just decoders.

Unfortunately, that's a bigger job, and it's not likely to be tackled soon, and this meets an immediate need, so I'm still inclined to merge it. Only two small changes:

* I think `print_decode_failures` (and the associate variable names) should be changed to `log_decode_failures` to keep the naming consistent with Go's log package and the LogMessage and LogError API.
* I'd like to see an entry in the "Backwards Incompatibilities" section of the changelog calling out that the DecoderRunner interface changed SetSendFailure to SetFailureHandling.

Thanks!

sathieu:
@rafrombrc Thanks for your review. Patch updated.

PR #1743, opened 2015-09-24
Add log_flags to hekad config

sathieu:


sathieu:
Complement #1742 for all the non-decoder errors.

Context: I need this because, under systemd, all stdout/stderr is logged (and forwarded to syslog, the timestamp is printed twice)

docs/source/config/index.rst
```diff
@@ -164,6 +164,13 @@ Config:
     The maximum size (in bytes) of message can be sent during processing.
     Defaults to 64KiB.
 
+.. versionadded:: 0.10
+
+- log_flags (int):
+    Control the prefix for STDOUT logs. Does not control the `LogOutput` plugin.
```
rafrombrc: STDOUT and STDERR

rafrombrc: Also, was there a reason why you didn't propagate the behaviour down to the LogOutput?

rafrombrc:
Needs changelog entry. :)

docs/source/config/index.rst
sathieu: No. Just that I don't need it.

i'll try to improve the patch.

PR #1751, opened 2015-09-29
added encoder for OpenTSDB

timurb:
Hi,
Here is an encoder for OpenTSDB.
It is loosely based on https://github.com/hynd/heka-tsutils-plugins/blob/master/lua_encoders/opentsdb_http.lua but is able to process full-formed JSON with depth of 1 level.
You'll probably have some suggestions on improvement prior to merging this which I'll be happy to fix.

### DIFF
--- a/sandbox/lua/encoders/opentsdb_http.lua
+++ b/sandbox/lua/encoders/opentsdb_http.lua
@@ -1,0 +1,190 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[=[
+Extracts Field data from messages and generates some OpenTSDB-compliant JSON
+(http://opentsdb.net/docs/build/html/api_http/put.html).
+
+Can optionally buffer and flush events after a configurable number of messages
+(flush_count) or after a configurable number of seconds have passed between the
+last Timestamp and the current one (flush_delta) (encoders currently have no
+support for timer_event()s)
+
+Config:
+
+- flush_count (number, default 1)
+    Flush the buffer every 'flush_count' messages.
+
+- flush_delta (number, default 0)
+    Flush the buffer if 'flush_delta' seconds have elapsed since the last
+    Timestamp and the current one.
+
+- tag_prefix (string, optional, default "")
+    Convert Fields to OpenTSDB tags if they match the tag_prefix.
+    (tag_prefix is stripped from the tag name)
+
+- tag_list (string, optional, default "")
+    A list of fields (separated by spaces) which will be converted into tags.
+    You can use either "tag_list", "tag_prefix" or both.
+
+- skip_fields (string, optional, default "")
+    A list of fields (separated by spaces) which will be dropped from processing.
+
+- type_as_prefix (boolean, optional, default true)
+    Prepend metrics when encoding them to OpenTSDB format by Type field.
+
+- ts_from_message (boolean, optional, default false)
+    Use the Timestamp field (otherwise "now")
+
+- add_hostname_if_missing (boolean, optional, default false)
+    If no 'host' tag has been seen, append one with the value of the
+    Hostname field.  Deprecated in favour of using the 'fieldfix' filter.
+
+
+*Example Heka Configuration*
+.. code-block:: ini
+
+    [OpentsdbEncoder]
+    type = "SandboxEncoder"
+    filename = "lua_encoders/opentsdb_http.lua"
+
+      [OpentsdbEncoder.config]
+      ts_from_message = true
+      add_hostname_if_missing = true
+      type_as_prefix = true
+      tag_prefix = "tag_"
+      tag_list = "env dc"
+      skip_fields = "Pid"
+
+    [opentsdb]
+    type = "HttpOutput"
+    message_matcher = "Fields[MetricOpentsdb] != NIL"
+    address = "http://opentsdb.example.com:4242/api/put"
+    encoder = "OpentsdbEncoder"
+
+*Example Output*
+.. code-block:: json
+    [{"value":0.05,"timestamp":1443535912,"metric":"stats.loadavg.15MinAvg","tags":{"env":"test","host":"example.com"}},{"value":2,"timestamp":1443535912,"metric":"stats.loadavg.NumProcesses","tags":{"env":"test","host":"example.com"}},{"value":0.01,"timestamp":1443535912,"metric":"stats.loadavg.5MinAvg","tags":{"env":"test","host":"example.com"}},{"value":0,"timestamp":1443535912,"metric":"stats.loadavg.1MinAvg","tags":{"env":"test","host":"example.com"}}]
+
+--]=]
+
+require "cjson"
+require "table"
+require "string"
+require "os"
+
+local l = require 'lpeg'
+l.locale(l)
+
+local flush_count     = read_config("flush_count") or 1
+local flush_delta     = read_config("flush_delta") or 0
+local tag_prefix      = read_config("tag_prefix")
+local tag_list_str    = read_config("tag_list") or ""
+local skip_fields_str = read_config("skip_fields") or ""
+local type_as_prefix  = read_config("type_as_prefix")
+local ts_from_message = read_config("ts_from_message")
+local add_hostname    = read_config("add_hostname_if_missing")
+_PRESERVATION_VERSION = read_config("preservation_version") or 1
+
+if not tag_prefix then
+  return -1
+end
+
+if type_as_prefix == nil then
+  type_as_prefix = true
+end
+
+if add_hostname == nil then
+  add_hostname = true
+end
+
+if ts_from_message == nil then
+  ts_from_message = true
+end
+
+tag_list = {}
+for field in tag_list_str:gmatch("[%S]+") do
+  tag_list[field] = true
+end
+
+skip_fields = {}
+for field in skip_fields_str:gmatch("[%S]+") do
+  skip_fields[field] = true
+end
+
+prefix = l.P(tag_prefix)
+chars = l.alnum + l.S("_-./")
+grammar = prefix * l.C(l.alpha * chars^0)
+
+buffer = {}
+last_flush = 0
+
+function flush()
+  add_to_payload(cjson.encode(buffer))
+  inject_payload()
+  last_flush = os.time()
+  buffer = {}
+end
+
+
+function process_message()
+
+  local ts
+  if ts_from_message then
+    ts = read_message("Timestamp") / 1e9
+  else
+    ts = os.time()
+  end
+
+  if not ts then return -1 end
+
+  local typ=read_message("Type")
+  local tags = {}
+  local metrics = {}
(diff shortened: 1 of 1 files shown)


trixpan:
@timurb great work. Looking forward for this to be merged. :+1:

sandbox/lua/encoders/opentsdb_http.lua
```diff
@@ -0,0 +1,190 @@
...
+
+- type_as_prefix (boolean, optional, default true)
+    Prepend metrics when encoding them to OpenTSDB format by Type field.
+
+- ts_from_message (boolean, optional, default false)
+    Use the Timestamp field (otherwise "now")
+
+- add_hostname_if_missing (boolean, optional, default false)
+    If no 'host' tag has been seen, append one with the value of the
+    Hostname field.  Deprecated in favour of using the 'fieldfix' filter.
```
rafrombrc: There is no `fieldfix` filter, this must be a reference to a custom filter that you're using internally?

sandbox/lua/encoders/opentsdb_http.lua
```diff
@@ -0,0 +1,190 @@
...
+local tag_list_str    = read_config("tag_list") or ""
+local skip_fields_str = read_config("skip_fields") or ""
+local type_as_prefix  = read_config("type_as_prefix")
+local ts_from_message = read_config("ts_from_message")
+local add_hostname    = read_config("add_hostname_if_missing")
+_PRESERVATION_VERSION = read_config("preservation_version") or 1
+
+if not tag_prefix then
+  return -1
+end
```
rafrombrc: You can't return here, you're not inside a function. At module scope, you can use the `error()` function to exit and return an error message. A better way to code this would be to remove L90-92 and update L82 to:

```lua
local tag_prefix = read_config("tag_prefix") or error("`tag_prefix` setting required")
```

sandbox/lua/encoders/opentsdb_http.lua
```diff
@@ -0,0 +1,190 @@
...
+
+skip_fields = {}
+for field in skip_fields_str:gmatch("[%S]+") do
+  skip_fields[field] = true
+end
+
+prefix = l.P(tag_prefix)
+chars = l.alnum + l.S("_-./")
+grammar = prefix * l.C(l.alpha * chars^0)
```
rafrombrc: Seems like these three vars should all be local.

sandbox/lua/encoders/opentsdb_http.lua
```diff
@@ -0,0 +1,190 @@
...
+    You can use either "tag_list", "tag_prefix" or both.
+
+- skip_fields (string, optional, default "")
+    A list of fields (separated by spaces) which will be dropped from processing.
+
+- type_as_prefix (boolean, optional, default true)
+    Prepend metrics when encoding them to OpenTSDB format by Type field.
+
+- ts_from_message (boolean, optional, default false)
+    Use the Timestamp field (otherwise "now")
```
rafrombrc: Seems weird to me that this would default to `false`. I'd think it more likely that folks will want to use the timestamp from the messages.

sandbox/lua/encoders/opentsdb_http.lua
```diff
@@ -0,0 +1,190 @@
...
+function process_message()
+
+  local ts
+  if ts_from_message then
+    ts = read_message("Timestamp") / 1e9
+  else
+    ts = os.time()
+  end
+
+  if not ts then return -1 end
```
rafrombrc: `Timestamp` is a static field, not a dynamic one, so there's no way it can return a `nil` value. Which means that AFAICT `ts` will never be nil, this line isn't needed.

sandbox/lua/encoders/opentsdb_http.lua
```diff
@@ -0,0 +1,190 @@
...
+
+  local typ=read_message("Type")
+  local tags = {}
+  local metrics = {}
+
+  local msg = decode_message(read_message("raw"))
+  if not msg.Fields then return -1 end
```
rafrombrc: When returning an error code it's a good idea to return an error message as the second return var, like so:

```lua
if not msg.Fields then
    return -1, "Malformed message (no fields)"
end
```

However, again, while a message might not have any fields, it should always have a `Fields` attribute, so I don't think this check is even necessary.

sandbox/lua/encoders/opentsdb_http.lua
```diff
@@ -0,0 +1,190 @@
...
+
+  for _, field in ipairs(msg.Fields) do
+    local name = field["name"]      -- these are related to way Heka works
+    local value = field["value"][1] --
```
rafrombrc: Lua idiom is to use dot notation when looking up static table keys, like so:

```lua
    local name = field.name
    local value = field.value[1]
```

sandbox/lua/encoders/opentsdb_http.lua
```diff
@@ -0,0 +1,190 @@
...
+
+    local tag = grammar:match(name) or (tag_list[name] ~= nil and name)
```
rafrombrc: Don't need the `~= nil` comparison.

sandbox/lua/encoders/opentsdb_http.lua
```diff
@@ -0,0 +1,190 @@
...
+    if tag then
+      tags[tag] = value
+    else
+      if type(value) == "number" and skip_fields[name] == nil then
```
rafrombrc: Nil comparisons are usually written using `not`, unless you explicitly want to differentiate btn `nil` and `false` values, so in this case you can use `if type(value) == "number" and not skip_fields[name] then`.

sandbox/lua/encoders/opentsdb_http.lua
```diff
@@ -0,0 +1,190 @@
...
+        if type_as_prefix then
+          name = typ .. "." .. name
```
rafrombrc: Our policy is to only use the `..` concatenation operator for single concatentations. Any time you'd use more than one `..`, you should use `string.format` instead.

sandbox/lua/encoders/opentsdb_http.lua
```diff
@@ -0,0 +1,190 @@
...
+        end
+        metrics[name] = value
+      end
+    end
+  end
+
+  if not tags["host"] and add_hostname then
+    tags["host"] = msg.Hostname
```
rafrombrc: `tags.host`

sandbox/lua/encoders/opentsdb_http.lua
```diff
@@ -0,0 +1,190 @@
...
+  end
+
+  opentsdb_message = {}
```
rafrombrc: This doesn't seem to be used anywhere, should be removed.

sandbox/lua/encoders/opentsdb_http.lua
```diff
@@ -0,0 +1,190 @@
...
+
+  for name, value in pairs(metrics) do
+    local message = {}
+    message["metric"] = name
+    message["value"] = value
+    message["timestamp"] = ts
+    message["tags"] = tags
+    buffer[#buffer+1] = message
```
rafrombrc: Believe it or not, Lua actually has to count the number of items in an array each time. Since we're adding metrics in a tight loop like this, it'd make sense to store and increment a separate variable for the buffer length, so it doesn't have to be recomputed each time around.

sandbox/lua/encoders/opentsdb_http.lua
```diff
@@ -0,0 +1,190 @@
...
+
+local flush_count     = read_config("flush_count") or 1
+local flush_delta     = read_config("flush_delta") or 0
+local tag_prefix      = read_config("tag_prefix")
+local tag_list_str    = read_config("tag_list") or ""
+local skip_fields_str = read_config("skip_fields") or ""
+local type_as_prefix  = read_config("type_as_prefix")
+local ts_from_message = read_config("ts_from_message")
+local add_hostname    = read_config("add_hostname_if_missing")
+_PRESERVATION_VERSION = read_config("preservation_version") or 1
```
rafrombrc: Our convention is to put the _PRESERVATION_VERSION stuff as the first lines of the module, immediately after the documentation comment. Also, you'll want to add in a value that you can increment in the source code, for times when changing the source code of the encoder invalidate the saved data. You can see an example of how this should be done in the [HTTP Status filter](https://github.com/mozilla-services/heka/blob/v0.10.0b1/sandbox/lua/filters/http_status.lua#L49).

rafrombrc:
Overall this looks great, thanks! I made a bunch of comments, but they're pretty much all for small stylistic issues.

In addition to the in-line comments, we'll want to have an entry added to the CHANGES.txt changelog, and the lovely documentation comment you've provided will need to be wired up to the `config` and `sandbox` sections of the Sphinx documentation.

sandbox/lua/encoders/opentsdb_http.lua
trink: Fields are optional and can be missing

rafrombrc: Okay, my bad, I thought read_message("raw") would always return at least an empty Fields table. In that case then just make sure to return an error message as specified above.

trink: Minimally you can get just a Uuid and a Timestamp (the required fields). Everything else can be nil.

sandbox/lua/encoders/opentsdb_http.lua
timurb: It was only in docs but really the default is true.
I'll update this and all other points shortly and update the pull request.

sandbox/lua/encoders/opentsdb_http.lua
timurb: This was left from @hynd's implementation. I'll remove the reference.

timurb:
The code is updated according to your feedback.
I've just realized that to avoid breaking @hynd's setup (whose original plugin I've taken to rework) I have to rename the lua file so it is now called `opentsdb_batch.lua`.
By the way should I put any credits for his implementation (from which I think only docs and `flush` function is left in place) and how to do that?

I've updated the docs but have never worked with RST so not very confident all is fine there, please let me know if it looks correct.

huhongbo:
+1

timurb:
Should I rewrite the code using `read_next_field()` in favor of `read_raw()` or `read_raw()` should be ok too?

I'm referring to this discussion: https://mail.mozilla.org/pipermail/heka/2015-October/000831.html
We were discussing the decoders and I wonder if anything from here is relevant for encoders too.

rafrombrc:
@timurb As I said in the mailing list thread, `read_message("raw")` is safe to use once the message has hit the router. Filter, output, and encode plugins all happen after the router, so `read_message("raw")` is what you want to use.

Also, using `tonumber` is fine, although if you break the `not skip_fields[name]` part into a separate if statement you can put that one first, so `tonumber` doesn't get called when the field is being skipped.

timurb:
Ok, I'll leave `read_message("raw")` here then.

Regarding `tonumber` -- if performance is fine let's leave this as is then. Checking for `not skip_fields[name]` first will add one more layer of `if`s which I believe makes the code harder to read.

PR #1752, opened 2015-10-01
Nginx error log have time

vlastv:


trink:
Let's just fix the grammar to put it in Timestamp.

vlastv:
I can fix grammar, but then need to correct all decoders apache and nginx.
Do it?

vlastv:
Grammar is described in a different repository, respectively, will be broken backward compatibility

trink:
That grammar is specific to nginx. As for breaking compatibility it will now work as expected and it clearly hasn't been used much (so I am not that concerned)

vlastv:
If the grammar should not be parsed into the heka schema then the PR is complete.
Otherwise, I have to change the grammar:
- time -> Timestamp
- pid -> Pid

Change the grammar?

trink:
We are talking about the nginx_error_grammar only: Pid is already correct and the Timestamp change we already discussed (that should be changed).

PR #1753, opened 2015-10-02
Update documentation use MariaDB Slow Query log

vlastv:
Slow query [begins with the time](https://github.com/mozilla-services/lua_sandbox/blob/dev/modules/mysql.lua#L57)

Updated settings according to the latest version of the documentation.


vlastv:
The only thing that I could not overcome, is that she is not sent to the last record until the new last.

trink:
Time is optional so this is incorrect

PR #1760, opened 2015-10-10
Add generic syslog decoder

sathieu:
(rsyslog coupled with software-specific extract)

First, it will decode according to rsyslog grammar (optional).

After, it will decode messages depending on it's programmname field.

Support for the following programs:
- PAM
- cron (CRON, crontab)
- isc-dhcp (dhclient, dhcpd)
- bind9 (named)
- OpenSSH (sshd)
- passwd (groupadd, groupdel, useradd)
- Postfix
- Puppet (puppet-agent)
- shadow (login)
- su
- sudo
- systemd (systemd-logind)

### DIFF
--- a/sandbox/lua/decoders/syslog.lua
+++ b/sandbox/lua/decoders/syslog.lua
@@ -1,0 +1,154 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[[
+Parses the syslog logs and extract fields of common programs.
+
+Config:
+
+- hostname_keep (boolean, defaults to false)
+    Always preserve the original 'Hostname' field set by Logstreamer's 'hostname' configuration setting.
+
+- rsyslog_template (string)
+    The 'template' configuration string from rsyslog.conf.
+    http://rsyslog-5-8-6-doc.neocities.org/rsyslog_conf_templates.html
+
+- tz (string, optional, defaults to UTC)
+    If your rsyslog timestamp field in the template does not carry zone offset information, you may set an offset
+    to be applied to your events here. Typically this would be used with the "Traditional" rsyslog formats.
+
+    Parsing is done by `Go <http://golang.org/pkg/time/#LoadLocation>`_, supports values of "UTC", "Local",
+    or a location name corresponding to a file in the IANA Time Zone database, e.g. "America/New_York".
+
+*Example Heka Configuration*
+
+.. code-block:: ini
+
+    [SyslogDecoder]
+    type = 'SandboxDecoder'
+    # Default 8MiB is not enough, use 16MiB
+    memory_limit = 16777216
+    filename = 'lua_decoders/syslog.lua'
+
+    [SyslogDecoder.config]
+    type = 'RSYSLOG_TraditionalFileFormat'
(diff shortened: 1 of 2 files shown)


sandbox/lua/decoders/syslog.lua
```diff
@@ -0,0 +1,182 @@
...
+            msg.Type = read_message('Type')
+        end
+        msg.Logger = read_message('Logger')
+        msg.Payload = read_message('Payload')
+        msg.EnvVersion = read_message('EnvVersion')
+        msg.Hostname = read_message('Hostname')
+        msg.Timestamp = read_message('Timestamp')
+        msg.Severity = read_message('Severity')
+        msg.Pid = read_message('Pid')
+        fields = read_message('Fields')
```
trink: Fields cannot be read this way

sathieu: @trink. Any alternative?

I expect having fields like _SYSTEMD_CGROUP=/user.slice/user-1000.slice/session-1.scope here.

If this is not possible, I will make rsyslog_template mandatory for now and we'll see later on for systemd-journal integration

timurb: @sathieu If you know fields names you can read them like `read_message("Fields[myfield]")`.
If you don't know field names there is a `read_next_field()` to iterate over all fields. See http://hekad.readthedocs.org/en/v0.9.2/sandbox/index.html for that.

Don't use `read_message('raw')` in decoder (as docs for v0.10.0b1 suggest) -- it is safe to use in filters and encoders but in decoders it sometimes works incorrectly, I've just hit this issue.

sathieu:
@trink What about the new patch?

trink:
I would require the grammar to be set and get rid of the else since read_next_field is deprecated.  (ideally the message model issues in the decoder would be addressed (from out of sync messages (mutation of conceptually immutable messages), to queuing and pool exhaustion)

sathieu:
I've choosen to only keep the programname dynamic field for now. Is it ok?

sathieu:
it now errors out when rsyslog_template is not defined or invalid.

trink:
As a future enhancement we may want to change the module library to allow addition of user defined prog_grammars and wildcard_grammars so custom apps and fallback grammars can be defined.

sathieu:
@trink, I like the idea, even if I don't know how to implement it.

I plan to add more prog_grammars, once my Heka setup has stabilized a bit (kernel/netfilter, proftpd, zorp, ctdb, heka itself, ...).

sathieu:
@trink, @rafrombrc Anything missing to merge it?

sandbox/lua/decoders/syslog.lua
```diff
@@ -0,0 +1,154 @@
...
+    if not hostname_keep then
+        msg.Hostname = fields.hostname or fields.source
+        fields.hostname = nil
+        fields.source = nil
+    end
+
+    msg.Payload = fields.msg
+    fields.msg = nil
+
+    -- Now parses syslog msg
```
trink, line 125: Couldn't this just be an option on the rsyslog decoder instead of duplicating the code here (defaulting to false to preserve the existing behaviour).

sathieu, line 125: It could, but the memory impact is not minor.

But the patch can be reworked to only require 'syslog_message' and 'postfix' when needed.

sathieu:
@trink @rafrombrc What can I do to move this PR forward?

sathieu:
Closing PR as Heka is a dead projet...

Issue #1757, opened 2015-10-08
Heka with the AMQP input plugin shuts down when the connection to the AMQP server drops


simonpasquier:
We've got Heka configured to read data from AMQP (RabbitMQ). With version 0.9, Heka was able to cope when the connection to the server was dropped but it looks like something is broken with 0.10.

Here is what happens when the AMQP server becomes unavailable:

```
2015/10/08 11:15:03 Pre-loading: [openstack_info_amqp]
2015/10/08 11:15:03 Pre-loading: [notification_decoder]
2015/10/08 11:15:03 Pre-loading: [HekaFramingSplitter]
2015/10/08 11:15:03 Loading: [HekaFramingSplitter]
2015/10/08 11:15:03 Pre-loading: [NullSplitter]
2015/10/08 11:15:03 Loading: [NullSplitter]
2015/10/08 11:15:03 Pre-loading: [ProtobufDecoder]
2015/10/08 11:15:03 Loading: [ProtobufDecoder]
2015/10/08 11:15:03 Pre-loading: [ProtobufEncoder]
2015/10/08 11:15:03 Loading: [ProtobufEncoder]
2015/10/08 11:15:03 Pre-loading: [TokenSplitter]
2015/10/08 11:15:03 Loading: [TokenSplitter]
2015/10/08 11:15:03 Loading: [notification_decoder]
2015/10/08 11:15:03 Loading: [openstack_info_amqp]
2015/10/08 11:15:03 Starting hekad...
2015/10/08 11:15:03 MessageRouter started.
2015/10/08 11:15:03 Input started: openstack_info_amqp
2015/10/08 11:21:06 Shutdown initiated.
2015/10/08 11:21:06 Stop message sent to input 'openstack_info_amqp'
2015/10/08 11:21:06 Waiting for decoders shutdown
2015/10/08 11:21:06 Stop message sent to decoder 'openstack_info_amqp-notification_decoder'
2015/10/08 11:21:06 Decoder 'openstack_info_amqp-notification_decoder': stopped
2015/10/08 11:21:06 Decoders shutdown complete
2015/10/08 11:21:06 Shutdown complete.
```

As @zstyblik noted in #1756, I suspect that this is more related to the AMQP library because the amqp_input.go code hasn't changed a lot between 0.9 and 0.10.

simonpasquier:
[...]

Issue #1756, opened 2015-10-07
error in AMQP output plugin when connection is reset


zstyblik:
Hello,

we've hit rather strange error with AMQP output plugin. Log output follows:

```
Sep 07 06:19:47 some-server hekad[9695]: 2015/09/07 06:19:47 Input started: JsonSocketInput
Oct 05 16:55:17 some-server hekad[9695]: 2015/10/05 16:55:17 Diagnostics: 32 packs have been idle more than 120 seconds.
Oct 05 16:55:17 some-server hekad[9695]: 2015/10/05 16:55:17 Diagnostics: (input) Plugin names and quantities found on idle packs:
Oct 05 16:55:17 some-server hekad[9695]: 2015/10/05 16:55:17 Diagnostics:         AMQPOutput: 32
Oct 05 16:55:17 some-server hekad[9695]: 2015/10/05 16:55:17
[...]
Oct 05 17:06:47 some-server hekad[9695]: 2015/10/05 17:06:47 Diagnostics: 32 packs have been idle more than 120 seconds.
Oct 05 17:06:47 some-server hekad[9695]: 2015/10/05 17:06:47 Diagnostics: (input) Plugin names and quantities found on idle packs:
Oct 05 17:06:47 some-server hekad[9695]: 2015/10/05 17:06:47 Diagnostics:         AMQPOutput: 32
Oct 05 17:06:47 some-server hekad[9695]: 2015/10/05 17:06:47
Oct 06 07:06:41 some-server hekad[9695]: 2015/10/06 07:06:41 Plugin 'AMQPOutput' error: write tcp AA.BB.CC.DD:5672: connection reset by peer
Oct 06 07:06:41 some-server hekad[9695]: 2015/10/06 07:06:41 Plugin 'AMQPOutput': stopped
Oct 06 07:06:42 some-server hekad[9695]: 2015/10/06 07:06:42 Plugin 'AMQPOutput': now restarting
Oct 06 07:06:42 some-server hekad[9695]: 2015/10/06 07:06:42 Plugin 'AMQPOutput' error: AMQP scheme must be either 'amqp://' or 'amqps://'
Oct 06 07:06:42 some-server hekad[9695]: 2015/10/06 07:06:42 Plugin 'AMQPOutput': now restarting
[...]
```

To be honest, I'm not sure whether the problem is with the plugin or AMQP library. I did look at the code and I don't think the problem is with AMQP plugin, but as I've said, I'm not sure. Please, advise.

[...]

PR #1764, opened 2015-10-14
Fix restart for the AMQP input plugin

simonpasquier:
This change modifies the Run() method to return something other than
nil when the connection to the AMQP server drops. This is partial fix
for bug #1757.

### DIFF
--- a/plugins/amqp/amqp_input.go
+++ b/plugins/amqp/amqp_input.go
@@ -175,7 +175,7 @@ func (ai *AMQPInput) Run(ir InputRunner, h PluginHelper) (err error) {
 	stream, err := ai.ch.Consume(ai.config.Queue, "", false, ai.config.QueueExclusive,
 		false, false, nil)
 	if err != nil {
-		return
+		return fmt.Errorf("Cannot consume from queue %s: %s", ai.config.Queue, err)
 	}
 
 	sRunner := ir.NewSplitterRunner("")
@@ -203,7 +203,8 @@ func (ai *AMQPInput) Run(ir InputRunner, h PluginHelper) (err error) {
 		}
 		msg.Ack(false)
 	}
-	return nil
+	// return an error message to trigger a potential restart of the plugin
+	return fmt.Errorf("Channel closed while reading from queue %s",  ai.config.Queue)
 }
 
 func (ai *AMQPInput) CleanupForRestart() {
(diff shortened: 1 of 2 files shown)


simonpasquier:
This doesn't fix completely the issue because now I see the same problem as reported by the issue #1756.

```
2015/10/14 08:57:22 Input 'openstack_info_amqp' error: Channel closed while reading from queue lma_notifications.info
2015/10/14 08:57:22 Input 'openstack_info_amqp': Restarting (attempt 1/-1)
2015/10/14 08:57:22 Input 'openstack_info_amqp' error: AMQP scheme must be either 'amqp://' or 'amqps://'
2015/10/14 08:57:23 Input 'openstack_info_amqp': Restarting (attempt 2/-1)
2015/10/14 08:57:23 Input 'openstack_info_amqp' error: AMQP scheme must be either 'amqp://' or 'amqps://'
2015/10/14 08:57:25 Input 'openstack_info_amqp': Restarting (attempt 3/-1)
2015/10/14 08:57:25 Input 'openstack_info_amqp' error: AMQP scheme must be either 'amqp://' or 'amqps://'
2015/10/14 08:57:27 Input 'openstack_info_amqp': Restarting (attempt 4/-1)
2015/10/14 08:57:27 Input 'openstack_info_amqp' error: AMQP scheme must be either 'amqp://' or 'amqps://'
```

plugins/amqp/amqp_input.go
```diff
@@ -203,7 +203,8 @@ func (ai *AMQPInput) Run(ir InputRunner, h PluginHelper) (err error) {
 		}
 		msg.Ack(false)
 	}
-	return nil
+	// return an error message to trigger a potential restart of the plugin
+	return fmt.Errorf("Channel closed while reading from queue %s",  ai.config.Queue)
```
rafrombrc: Wouldn't this return an error even during normal shutdown? We need a way to distinguish btn the AMQP channel closing unexpectedly and it closing as expected b/c the `Stop` method was called.

simonpasquier: I'll do more tests but it didn't seem to be an issue during my initial testing.

simonpasquier: So I tested again. Indeed the Stop() method will close the AMQP channel and the Run() method will return an error but the plugin runner detects that Heka is shutting down and it doesn't try to respawn the plugin [0].

[0] https://github.com/mozilla-services/heka/blob/versions/0.10/pipeline/plugin_runners.go#L328

rafrombrc: That's fine, but the purist in me is still a bit put off by returning an error when there isn't really an error case. I'll add some code so the `Stop` method sets a flag, which the `Run` method can check on the way out. If the flag is set, we'll return nil; if not, we return the error.

PR #1770, opened 2015-10-22
Use io.ReadFull to prevent situation where gzip.Reader doesn't fill the buffer

pib:
I ran into this issue trying to ship some old log files. I traced it to this read. The issue is that the gzip.Reader was not returning the full `LINEBUFFERLEN` bytes, but also not returning an error. I'm assuming it has something to do with blocks inside of gzip files.

It was resulting in the same file being considered to be a new file which caused the LogStreamerInput to jump back to the beginning of the file and start reading it again, over and over.

This issue is still present in 0.10 as well.

### DIFF
--- a/logstreamer/reader.go
+++ b/logstreamer/reader.go
@@ -422,7 +422,7 @@ func SeekInFile(path string, position *LogstreamLocation) (*os.File, io.Reader, 
 		expectedN int64
 	)
 	if seekPos >= 0 {
-		n, err = reader.Read(buf)
+		n, err = io.ReadFull(reader, buf)
 		expectedN = int64(LINEBUFFERLEN)
 	} else {
 		n, err = reader.Read(buf[-seekPos:])


pib:
I don't understand how this patch could have caused tests to not pass. io.ReadFull does the same thing as Reader.Read except it makes sure it fills the buffer.

sathieu:
@plb Heka tests sometimes fail on travis (I can reproduce this sometimes using docker on my local machine). Run the tests on your local machine to be sure.

pib:
Unfortunately,  on my local machine even more tests fail :D 
I'm guessing the Heka developers don't test on OSX too often?
I can try it in a Linux VM later and see if it completes there.

My only guess as to why this would legitimately fail is that it may not return an EOF in some situations when a regular Read call would. But even then, the next read should result in an EOF being returned.

rafrombrc:
A few notes:

* We still have a few tests that intermittently fail on travis, but there have been some fixes on the 0.10 branch that reduce the number greatly
* I know the tests were passing on OSX when I was doing 0.9 development, at least before releases. I did, however, verify that the 0.9 branch's sandbox tests seem to be failing on OSX right now due to linking issues. However, 0.9 is no longer under active development, so this isn't a high priority. 0.10 branch tests are all passing for me on OSX.
* I've checked out your branch and verified that the tests are passing on Linux.
* I'm happy to merge this, but there are no plans at this point for another 0.9 release. Are you willing to close this PR and open another one against the versions/0.10 branch? It should also include an entry in the 'bug handling' section of the CHANGES.txt changelog for the 0.10 release.

Thanks so much!

pib:
@rafrombrc Sure, I can do a PR against 0.10 instead. I've already made a release on the OwnLocal fork with the change and pointed our build system at that so the fix is available for us to use until we switch to 0.10 later.

PR #1775, opened 2015-10-24
Add Fluent protocol decoder and encoder

salekseev:
This supersedes https://github.com/mozilla-services/heka/pull/1666

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,7 +48,9 @@ endif()
 if (INCLUDE_DOCKER_PLUGINS)
     message(STATUS "Docker plugins enabled.")
     set(PLUGIN_LOADER ${PLUGIN_LOADER} "github.com/mozilla-services/heka/plugins/docker")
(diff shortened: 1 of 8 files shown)


rafrombrc:
First, thanks so much for your contribution, and I'm very sorry to have taken so long before looking at it.

Unfortunately, I'm not going to accept this PR for merging. We very much try to discourage decoders from being written in Go. I'd be much more interested in a SandboxDecoder that provided similar functionality, using either https://github.com/antirez/lua-cmsgpack or one of the native Lua msgpack implementations to do the heavy lifting. Even better would be an implementation that provided a pure Lua implementation, but which would make use of the much faster C-backed version if the module was available.

Because it's not possible at this time to implement a splitter in Lua, I'd be willing to consider merging a separate PR that only included the MsgPack splitter, but it would need to also include documentation and an entry in the CHANGES.txt changelog.

Thanks again.

Issue #1750, opened 2015-09-26
Kafka Output does honour Sarama Circuit Break errors as RetryMessageError


trixpan:
@trink 

During some chaos monkey testing I have noticed that Kafka Output will not fully interpret Sarama AsyncProducer delivery errors, instead, receiving the errors via the appropriate channel but by then the  message is gone out of the pipeline.

As consequence, messages will be lost during for example a full cluster broker outage (even when buffering is turned on).

To reproduce:

Relevant hekad config section:

```
[TestKafkaOutput]
type = "KafkaOutput"
message_matcher = "Type == 'TcpInput'"
topic = "test_topic"
addrs = ["localhost:9092"]
encoder = "PayloadEncoder"
use_buffering = true
    [TestKafkaOutput.buffering]
    max_file_size = 268435456
    max_buffer_size = 1073741824
    full_action = "block"
    cursor_update_count = 1
```

set up a single node kafka cluster, start hekad, start injecting events:

In parallel, stop the kafka broker process (kill or ctl-c). 

Heka will generate the following sequence of messages:

```
2015/09/25 23:12:12 Plugin 'TestKafkaOutput' error: kafka error 'kafka: client has run out of available brokers to talk to (Is your cluster reachable?)' for message '<message_contents>'
2015/09/25 23:12:13 Plugin 'TestKafkaOutput' error: kafka error 'kafka: client has run out of available brokers to talk to (Is your cluster reachable?)' for message '<message_contents>'
2015/09/25 23:12:14 Plugin 'TestKafkaOutput' error: kafka error 'circuit breaker is open' for message '<message_contents>'
```

Not sure how that could be done asynchronously but perhaps Sarama errors should trigger Heka's [RetryMessageError](http://hekad.readthedocs.org/en/v0.10.0b1/developing/plugin.html#retry-message-error) ?

Issue #1749, opened 2015-09-26
Enabling buffering on Kafka output lead leads to infinite repetition of messages upon restart


trixpan:
Configuring buffering on a Kafka Output will lead to heka to send all messages again whenever heka is restarted.

```
[TestKafkaOutput]
type = "KafkaOutput"
message_matcher = "Type == 'TcpInput'"
topic = "test_topic"
addrs = ["localhost:9092"]
encoder = "PayloadEncoder"
use_buffering = true
    [TestKafkaOutput.buffering]
    max_file_size = 268435456
    max_buffer_size = 1073741824
    full_action = "block"
    cursor_update_count = 1
```

The reason seem to be the fact that all messages get inserted into the heka buffers but none of them seems to ever come out. 

As consequence upon every single reboot all log messages are resent but the backlogs never cleared.

xrl:
What version of heka are you running? `0.10b` has some queueing bugs which I believe are in the process of being sorted out.

trixpan:
@xrl  

I run the dev version. The issue is not related to buffering per-se but to the way Sarama is called without this resulting on a cursor update.

Note how Kafka's output 

https://github.com/mozilla-services/heka/blob/dev/plugins/kafka/kafka_output.go#L375

differs from 

https://github.com/mozilla-services/heka/blob/dev/plugins/http/http_output.go#L117

rafrombrc:
Fixed in #1887.

PR #1777, opened 2015-10-28
Converts Kafka output to Synchronous publisher 

trixpan:
* Converts Kafka publisher to synchronous publisher resulting in better control of delivery failures; #1750
* Introduces buffering to Heka's kafka at the expense of some performance. #1749

* Update the documentation to reflect drop of the Mercurial requirements ##1736

### DIFF
--- a/cmake/externals.cmake
+++ b/cmake/externals.cmake
@@ -19,7 +19,7 @@ if(INCLUDE_SANDBOX)
     externalproject_add(
         ${SANDBOX_PACKAGE}
         GIT_REPOSITORY https://github.com/mozilla-services/lua_sandbox.git
-        GIT_TAG 82ae00d85f3111ee77bf15dbcb97264f177a66ec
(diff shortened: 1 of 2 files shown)


trixpan:
@trink @rafrombrc is the unbuffered message loss issues identified as part of #1750 and #1749 worth this PR ?

rafrombrc:
First, thanks so much for your contribution, and I'm sorry it's taken me so long to get around to looking at it. Things have gotten much busier for me, and my time to work on Heka core has diminished lately.

With that being said, I'm afraid I can't accept this PR. The performance impact of changing to synchronous delivery is too high, and this would greatly interfere with the real world requirements we have for Heka/Kafka throughput in Mozilla. I might be willing to consider a PR that allows synchronous delivery as a configurable option, with async delivery being the default, but removing async delivery altogether is not a viable option.

PR #1784, opened 2015-11-09
RstEncoder: Fix Timestamp value and change it to Local format

yunkai:
- The value of Message.GetTimestamp() is seconds not nanoseconds,
  it should be passed into time.Unix(sec, nsec) as first argument.

- Convert timestamp to Local format instead of UTC will be more
  user friendly.

Signed-off-by: Yunkai Zhang <qiushu.zyk@taobao.com>


yunkai:
Sorry, I misunderstand the unit of Message.Timestamp.

PR #1789, opened 2015-11-17
Add CloseAfterWrite option to the TcpOutput

TheDude05:
This PR adds an option to the TcpOuput plugin that closes the socket after successfully writing a message. This is to deal with TCP servers that close their sockets without SO_LINGER enabled.

I discovered this issue while trying to write a JSON payload to Sensu which closes its side of the TCP connection after reading up to a newline. When that happens, the TCP socket on the Heka side transitions to CLOSE_WAIT and is stuck there until we try and write another message. Here is an example:

```
# lsof -i:3030 -P -n
COMMAND     PID  USER   FD   TYPE    DEVICE SIZE/OFF NODE NAME
sensu-cli  4809 sensu   10u  IPv4  85956468      0t0  TCP 127.0.0.1:3030 (LISTEN)
sensu-cli  4809 sensu   11u  IPv4  85956469      0t0  UDP 127.0.0.1:3030 
hekad     11504  root   21u  IPv4 203961716      0t0  TCP 127.0.0.1:43054->127.0.0.1:3030 (CLOSE_WAIT)
```

Unfortunately (at least on Linux) the kernel will still accept writes on the file descriptor despite the TCP socket being in CLOSE_WAIT. The result is that when we write that second message to this closed TCP socket, it will not be delivered because the remote end is gone. This particular issue is similar to the issue outlined here: https://github.com/golang/go/issues/10940

I've occasionally seen this second message be delivered after restarting Heka but I have not been able to consistently reproduce that behavior.

With this PR, you can enable 'close_after_write' on the TcpOutput plugin which closes the socket and avoids this CLOSE_WAIT issue.

### DIFF
--- a/docs/source/config/outputs/tcp.rst
+++ b/docs/source/config/outputs/tcp.rst
@@ -55,6 +55,11 @@ Config:
     All of the :ref:`buffering <buffering>` config options are set to the
     standard default options, except for `cursor_update_count`, which is set to
     50 instead of the standard default of 1.
+- close_after_write (bool, optional):
+    Specifies whether or not the TCP socket should be closed after Heka
+    successfully writes a message. This option exists to deal with buggy TCP
+    servers that close their socket before our Heka client does. This should
+    only be enabled if you absolutely need it. Defaults to false.
 
 Example:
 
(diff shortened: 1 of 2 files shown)


rafrombrc:
Oh, sheesh, I merged this, forgetting that a more comprehensive feature that covers this use case had already been merged to the dev branch (see #1667). I'll revert this, and will cherry-pick #1667 so that the fix shows up in the versions/0.10 release.

TheDude05:
That sounds great. Sorry for the noise as I hadn't seen that TCP re-establishment issue. It looks like that should handle our use case.

PR #1793, opened 2015-11-19
Bugfix/elasticsearch output

tianchaijz:
- Bugfix: The `ProcessMessage` function in the ES plugin is not synchronized (the `append` function is not thread safe) when goroutines running in parallel.
- Bugfix: Should Drop empty batch instead of sending it to ES (ES will respond 400).
- Enhancement: Only log error message when ES's resonpse code != 200.

 But I didn't find the source code that makes `ProcessMessage` function run in multiple goroutines?

### DIFF
--- a/plugins/elasticsearch/elasticsearch.go
+++ b/plugins/elasticsearch/elasticsearch.go
@@ -381,6 +381,10 @@ func (h *HttpBulkIndexer) Index(body []byte) (err error, retry bool) {
 	var response_body []byte
 	var response_body_json map[string]interface{}
 
+	if len(body) == 0 {
+		return nil, false
+	}
+
 	url := fmt.Sprintf("%s://%s%s%s", h.Protocol, h.Domain, h.Path, "/_bulk")
 
 	// Creating ElasticSearch Bulk HTTP request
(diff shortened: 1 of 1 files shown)


rafrombrc:
This code looks great, it's definitely easier to follow than the use of the `flushManual` channel with the weird blocking in it that I had in there originally. I was trying to avoid the need for a channel put for every single message, but I think that the simplification of the code is probably worth it. I don't suppose you've done any throughput comparisons from before and after this change?

A couple of notes:
* I think `recvChan` should just be `chan MsgPack`, not `chan *MsgPack`. MsgPacks are very small, so there's not a big cost in mem copies, and if we don't use pointers then we're guaranteed that all MsgPack allocation will be on the stack, avoiding any garbage collection.
* This fix should be landing on the versions/0.10 branch instead of dev, so we can make sure it gets into the v0.10.0b2 release.

 I am a bit concerned about the original issue, however. The idea is that ProcessMessage should *not* be getting called from more than one goroutine. Any plugin should have either ProcessMessage or TimerEvent called at any one time, never both, and never any one of them multiple concurrent times. If ProcessMessage is in fact being invoked from multiple separate goroutines, then that bug needs to be found and fixed or else there will be additional problems down the road. I'll poke around and see if I can find any place that might be happening. Any help you can offer tracking down the issue would be appreciated.

tianchaijz:
Hi Rob,

Thanks for your suggestions, I'll follow them. I haven't done any throughput comparisons, the monitor of ES shows that the indexing rate is double however, since there are no damaged bulks sent to ES.

tianchaijz:
Hi guluglu,

Are your messages encoded in UTF-8? If not, they will be dropped by ES, it can be found in ES's log.

rafrombrc:
Closed, see #1795.

PR #1795, opened 2015-11-20
Bugfix/elasticsearch plugin

tianchaijz:
#1793


rafrombrc:
Okay, I did some further testing and verified that the problem is *not* that ProcessMessage is being called multiple times. This is a relief, because that would point to much bigger issues.

What's happening is that there's a (obvious in retrospect) race condition where the flushTicker might fire while ProcessMessage is running. The `flushManual` hack prevents both flushes from happening at the same time, but it doesn't prevent a flushTicker flush from happening concurrently with an outBatch array append, leading to the corrupted records. Your change to the code resolves this issue.

I'll go ahead and get this merged tonight. Thanks for your help.

PR #1802, opened 2015-12-01
Move debian directory out of root directory

sathieu:
This is a reserved name


sathieu:
See also https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=805278#20

sathieu:
heka.service.in should probably be moved directly in packaging/ as it is not Debian specific. But I've not found how yet (and configure_file is not needed for this file).

sathieu:
superseeded by #1804

PR #1816, opened 2015-12-18
Ts line protocol cleanup

rafrombrc:


davidbirdsong:
a field whitelist would be very useful. from my reading, the magic values
get you close, but i didnt see a way to block all fields and only add
explicit without the single value option.

Issue #1801, opened 2015-11-25
LogstreamerInput read current postion in file


guluglu:
LogstreamerInput can read the newest file in a directory.But when there are many machine,the newest file is about 2~3g,the total amount is still too large.Is there any  way to read from the current postion in file not from the begin

guluglu:
Like start_position(string, one of ["beginning", "end"]) setting in Logstash

tclh123:
I need this feature too. related #1482
cc @rafrombrc

rafrombrc:
Fixed in #1831. Thanks so much to @tclh123 for submitting the code. :)

guluglu:
@rafrombrc  Hi, I test it, but it doesn't work when file is small than 500 bytes i.e. it  still read from begin though I set initial_tail as true

tclh123:
@guluglu should be fixed in #1838

PR #1831, opened 2016-01-12
support initial_tail option in LogstreamerInput

tclh123:
feature request by #1482, #1801
some code based on #1486.


tclh123:
According to `SeekInFile`，maybe I should deal with gzipped file in `SetToTail` too.

PR #1832, opened 2016-01-12
Add websockets output

waqark3389:
Add websockets output


rafrombrc:
I'm afraid this isn't the right way to get a new plugin into the Heka core. What you'll want to do is to actually integrate your plugin into the Heka code tree and open from your repo to this repo that shows all of the code that your repo contains. Here's an example of a PR that introduces a new Go plugin into the Heka code base:

https://github.com/mozilla-services/heka/pull/1491

That one does a pretty good job, in that it updates the documentation and the changelog in addition to providing the implementation. It doesn't have any tests, however, which would be even better.

Please compare that PR to this one, hopefully that will help you see how to approach this differently.

Thanks for your efforts!

PR #1842, opened 2016-01-26
Implementation of a multiline splitter

relistan:
We work with a lot of Java and Scala stacktraces and the other options I've tried for supporting them in Heka don't work as well as I'd like. This is an implementation of a regex-based `MultilineSplitter` which works great for our stacktraces. The implementation is that you define a regex to use as the delimiter and a regex used to match lines that should be joined together. It first splits the buffer using the delimiter and then checks each section against the multiline regex to see if it's a match. All lines that are contiguous and match the multiline regex are joined. Because of the multiline nature, it always keeps the delimiter on the EOL.

This is going to be notably slower than the `RegexSplitter` because it has to find many matches on the first pass rather than the first one. (That's limited to 99 by default and is not currently configurable without a recompile.) Secondly, it will run a second regex on all those matches. Given's Go's performance-oriented Regex implementation and reasonable logging levels it appears to be tolerable. It can, in the worst case, re-split the first lines in a very large buffer repeatedly.

Here's an example configuration for the splitter:
```toml
[multiline_splitter]
type = "MultilineSplitter"
multiline = '(\] FATAL )|(\A\s*.+Exception: .)|(at \S+\(\S+\))|(\A\s+... \d+ more)|(\A\s*Caused by:.)|(\A\s*Grave:)'
delimiter = '\n'
```
Given a broken Kafka installation, this generates something like the following when encoded with the `ESLogstashV0Encoder`:
```json
{
  "@fields": {
    "ContainerName": "boring_bohr",
    "ContainerID": "910f097243d6"
  },
  "@source_host": "docker1",
  "@uuid": "bda1cb47-8aa4-420e-9316-543364afd5fc",
  "@timestamp": "2016-01-24T14:22:17",
  "@type": "message",
  "@logger": "stdout",
  "@severity": 7,
  "@message": "java.net.ConnectException: Connection refused\n\tat sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)\n\tat sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:716)\n\tat org.apache.zookeeper.ClientCnxnSocketNIO.doTransport(ClientCnxnSocketNIO.java:361)\n\tat org.apache.zookeeper.ClientCnxn$SendThread.run(ClientCnxn.java:1081)\n",
  "@envversion": "",
  "@pid": 0
}
```
Note that this output is from a splitter-enabled Docker input plugin that I will prepare a separate PR for.

### DIFF
--- a/pipeline/splitters.go
+++ b/pipeline/splitters.go
@@ -23,9 +23,14 @@ import (
 	"crypto/subtle"
 	"errors"
 	"fmt"
-	"github.com/mozilla-services/heka/message"
 	"hash"
 	"regexp"
+
+	"github.com/mozilla-services/heka/message"
+)
+
+const (
+	MULTILINE_SPLITTER_MAX_LINES = 99
 )
 
 type NullSplitter struct {
@@ -92,6 +97,78 @@ func (t *TokenSplitter) FindRecord(buf []byte) (bytesRead int, record []byte) {
 			}
 			bytesRead += n + 1
 		}
+	}
+
+	return bytesRead, buf[:bytesRead]
+}
+
+type MultilineSplitter struct {
+	delimiter  *regexp.Regexp
+	multiline  *regexp.Regexp
+	captureLen int
+}
+
+type MultilineSplitterConfig struct {
+	Delimiter string
+	Multiline string
+}
+
+func (r *MultilineSplitter) ConfigStruct() interface{} {
+	return &MultilineSplitterConfig{
+		Delimiter: "\n",
+	}
+}
+
+func (r *MultilineSplitter) Init(config interface{}) error {
+	conf := config.(*MultilineSplitterConfig)
+	var err error
+	if r.delimiter, err = regexp.Compile(conf.Delimiter); err != nil {
+		return err
+	}
+	if r.delimiter.NumSubexp() > 1 {
+		return fmt.Errorf("regex must not contain more than one capture group: %s",
+			conf.Delimiter)
+	}
+
+	if r.multiline, err = regexp.Compile(conf.Multiline); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *MultilineSplitter) FindRecord(buf []byte) (bytesRead int, record []byte) {
+	var loc [][]int
+	loc = r.delimiter.FindAllSubmatchIndex(buf[r.captureLen:], MULTILINE_SPLITTER_MAX_LINES) // Up to 99 matches
+	if loc == nil {
+		return 0, nil
+	}
+
+	// Check the first one
+	if !r.multiline.Match(buf[:loc[0][1]]) {
+		bytesRead = loc[0][1]
+		return bytesRead, buf[:bytesRead]
+	}
+
+	// Loop through, looking for the first delimiter not also matching a multiline
+	var lastDelimiter []int
+	previous := []int{ 0, 0 }
+	for _, lastDelimiter = range loc[1:] {
+		// Check for a multiline match inside this record
+		if r.multiline.Match(buf[previous[1]:lastDelimiter[0]]) {
+			previous = lastDelimiter // Set to the position of previous delimiter
+			continue
+		}
+
+		// We didn't match, so fall back one
+		lastDelimiter = previous
+		break
+	}
+
+
+	// We always keep the delimiter at EOL
+	bytesRead = lastDelimiter[1]
+	if bytesRead == 0 {
+		return 0, nil
 	}
 
 	return bytesRead, buf[:bytesRead]
@@ -282,6 +359,9 @@ func init() {
 	})
 	RegisterPlugin("RegexSplitter", func() interface{} {
 		return &RegexSplitter{}
+	})
+	RegisterPlugin("MultilineSplitter", func() interface{} {
+		return &MultilineSplitter{}
 	})
 	RegisterPlugin("HekaFramingSplitter", func() interface{} {
 		return &HekaFramingSplitter{}
--- a/pipeline/splitters.go
+++ b/pipeline/splitters.go
@@ -149,6 +149,15 @@ func (r *MultilineSplitter) FindRecord(buf []byte) (bytesRead int, record []byte
 		return bytesRead, buf[:bytesRead]
 	}
 
+	if len(loc) == 1 {
+		// In this scenario we are on a multiline but missed the full record in this
+		// read, so we just emit what we have. It would be nice to just try this again on
+		// next read but without keeping more state, we could try forever if somehow that
+		// next line never comes (i.e. truncated input).
+		bytesRead = loc[0][1]
+		return bytesRead, buf[:bytesRead]
+	}
+
 	// Loop through, looking for the first delimiter not also matching a multiline
 	var lastDelimiter []int
 	previous := []int{ 0, 0 }
--- a/pipeline/splitters.go
+++ b/pipeline/splitters.go
@@ -98,34 +98,34 @@ func (t *TokenSplitter) FindRecord(buf []byte) (bytesRead int, record []byte) {
 	return bytesRead, buf[:bytesRead]
 }
 
-type MultilineSplitter struct {
+type PatternGroupingSplitter struct {
 	delimiter *regexp.Regexp
-	multiline *regexp.Regexp
+	grouping *regexp.Regexp
 	maxLines  int
 }
 
-type MultilineSplitterConfig struct {
+type PatternGroupingSplitterConfig struct {
 	Delimiter string
-	Multiline string
+	Grouping string
 	MaxLines  int `toml:"max_lines"` // Maximum lines to group together
 }
 
-func (r *MultilineSplitter) ConfigStruct() interface{} {
-	return &MultilineSplitterConfig{
+func (r *PatternGroupingSplitter) ConfigStruct() interface{} {
+	return &PatternGroupingSplitterConfig{
 		Delimiter: "\n",
 		MaxLines:  99,
 	}
 }
 
-func (r *MultilineSplitter) Init(config interface{}) error {
-	conf := config.(*MultilineSplitterConfig)
+func (r *PatternGroupingSplitter) Init(config interface{}) error {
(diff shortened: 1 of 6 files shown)


rafrombrc:
I'm wary of this one, especially since it's named "MultilineSplitter" when it may or may not be the right choice for all multi-line splitting. You say that the "other options for supporting them don't work as well as [you]'d like". I assume that the RegexSplitter is the other option to which you're referring. Can you explain exactly what it is about the RegexSplitter that's causing you pain, and how this splitter resolves the issue?

Thanks!

relistan:
I'm totally open to naming it something else, that's not an issue for me at all. I didn't want to call it `StacktraceSplitter` because it's actually good for most multi-line things as far as I can tell. It works a lot like, but not exactly like, multi-line grouping in other logging frameworks.

As you say the only other option for this is the `RegexSplitter` but:

1. It can only match on the beginning or end of a line which is not good enough in many cases to catch something that should be grouped multi-line.
1. It usually misses the first line in a stacktrace, the one with the error message,
1. It doesn't work easily with lots of stacktraces, e.g. stacktraces that get emitted without a leading space on each line, because:
1. It has a limitation of a single capture group, which I assume is to simplify the code and for performance reasons: https://github.com/mozilla-services/heka/blob/dev/pipeline/splitters.go#L125 . It would be more usable for this purpose without that limitation. I was not enthusiastic about changing the behavior of that splitter.

Do you have specific reservations that I might be able to address?

rafrombrc:
I don't have any specific comments on the implementation yet, I haven't dug in; I'm working my way through your DockerLogInput overhaul first. In general, though, I'm averse to having two splitter implementations that are so similar to each other. If at all reasonably possible, I'd much rather find a way to combine them into a single splitter which provides the functionality you need but which doesn't sacrifice any performance for the simpler cases already supported by the RegexSplitter. When I do get to reviewing this PR, I'll be doing so with an eye for how we might merge the two.

relistan:
Thanks Rob, sounds reasonable. I could take a stab at a combined splitter if that's your preference. It's easy enough to have some configuration options that support everything the `RegexSplitter` does now (and rename this splitter to `RegexSplitter`). From a separation of concerns standpoint this separate splitter seemed like the right thing to do, but maybe from a user confusion standpoint of view combining them is better. I probably won't have time until early next week to work on it, though.

BTW saw your presentation on Heka at the first GopherCon, which is how we ended up running it at New Relic for the last year and a half or so. :+1:

rafrombrc:
I've looked at this (and your explanation for this) a little bit more deeply, and I still don't understand what this provides that the RegexSplitter doesn't. Either I'm not understanding what your splitter provides, or you're not fully understanding the RegexSplitter behaviour, not sure which. I'll respond to your 4 points here:
1. Despite the perhaps poorly named `delimiter_eol` setting, RegexSplitter doesn't match only on the beginning or end of a line, it matches the beginning or end of a _record_. The regex you use can match multiple lines. The regex can match the entire record, if need be, as long as you use a match group to make sure that what you want to keep is included in the output record.
2. I don't understand this. The record generated by the RegexSplitter will contain everything not matched by the provided regular expression plus the contents of any match group in the regular expression, with the match group data at either the beginning or end of the record depending on the `delimiter_eol` value. How does it miss the first line in a stacktrace?
3. I also don't understand this. What do leading characters (or lack thereof) have to do with finding the text that denotes a record boundary?
4. The purpose of the match group is specify what part, if any, of the matched regular expression should be included in the resulting record. If your regular expression doesn't contain a match group, then the text matched by the regex will be dropped from the resulting record. If it _does_ contain a match group, then the text matched by the group will be included in the resulting record. All text in between the matched delimiter will _always_ be included in the resulting record. I suppose we could add support for multiple match groups here, but as it is you can always put a match group around your entire regex to make sure you get all of it, and then parse out the information that you care about in the decoding layer.

To restate, the existing RegexSplitter is meant to let you define a regular expression that will match the minimum identifiable record boundary. The generated record will contain _all_ of the data in between those boundaries, however, not just what was matched by the regular expression. In fact, the regular expression match won't even be included, unless you specify a match group.

Hopefully that clarifies the behaviour of the existing splitter. But maybe you already understood that, and your implementation meets some use cases that RegexSplitter doesn't. If that's so, I'd love to get a more detailed explanation of exactly what those are, preferably with a set of records that I can examine for which the RegexSplitter is not sufficient.

relistan:
Thanks for the explanation. But, I've been through this code a bunch and I guess I'm just totally not understanding this. I don't see how you can capture full stacktraces that are intermixed with other logs. I did say "line" but I'm referring to getting a stacktrace into a single record, while not also messing up all the other logging that isn't stacktraces.

If you wouldn't mind, please take a look at the [tests I wrote for the `MultilineSplitter`](https://github.com/mozilla-services/heka/pull/1842/files#diff-507c8b89ca8d3be322d385790bfe92f9R159). What can I feed to the `RegexSplitter` that will capture all of that? And also work for non-stacktrace lines like in the tests? Are you guys doing this? If so, what expression are you using? Googling many times before, and again now, was unsuccessful in finding anyone doing this publicly.

As far as I can tell, the code is only going to capture the leftmost match using `FindSubmatchIndex` and there is no `for` loop. And it always returns a record unless it didn't match at all, in which case it also returns 0 bytes read. Doesn't that mean I can only ever match the first match?

I'm good with regexes, but maybe I'm just missing something that makes this all just work.
Here's a [Golang Playground](https://play.golang.org/p/BqpoX2PVKE) demonstrating some attempts to make a match work with `FindSubMatchIndex` against one of the test cases.

rafrombrc:
Okay, now I understand the disconnect... it seems daft in retrospect, but I didn't keen on to the fact that you're trying to capture records of different types, i.e. multiline tracebacks embedded _within_ log records that are only a single line.

If there were some way to get the framework to emit a timestamp (or other consistent piece of text) before _every_ record, then RegexSplitter will work, it doesn't care about what's in the records, just that it found a delimiter. But I realize that might be hard with tracebacks, since they're generated by error conditions, so you probably don't get much control over what they look like.

I have a few more thoughts on this, but have to go into a meeting now, more soon.

relistan:
Ah, that explains the disconnect! Glad we're on the same page. Thanks!

relistan:
Any further thoughts on this, @rafrombrc ?

rafrombrc:
Sorry for the delay, my time for Heka stuff is pretty minimal these days.

My first thought is that my preference, from an end user perspective, would be to add support to the TokenSplitter and RegexSplitter to add an optional 'traceback' regex. Those splitters would behave exactly as they do now (w/ no performance loss) if there was no traceback regex. If one was included, they would use it to check for tracebacks along the way.

That might be more work than you're up for, though. I'm willing to include this splitter, but would want it to be renamed. MultilineSplitter is confusing, b/c it implies you need this to split any input with multiline records... that's part of what threw me, anyway. MultiPatternSplitter, maybe? TracebackSplitter? Anyway, there are a few other things that would also need to happen:
- Add documentation to our Sphinx in the `docs` directory, clearly describing the use case and the performance trade-offs.
- Add an entry to the 'features' section for 0.11 in the CHANGES.txt changelog.

I'll also make some in-line comments, am reviewing the code now.

pipeline/splitters.go
```diff
@@ -23,9 +23,14 @@ import (
...
 	"fmt"
-	"github.com/mozilla-services/heka/message"
 	"hash"
 	"regexp"
+
+	"github.com/mozilla-services/heka/message"
+)
+
+const (
+	MULTILINE_SPLITTER_MAX_LINES = 99
```
rafrombrc: Seems like this should be a config option on the plugin rather than a hard-coded constant.

pipeline/splitters.go
```diff
@@ -97,6 +102,87 @@ func (t *TokenSplitter) FindRecord(buf []byte) (bytesRead int, record []byte) {
...
+func (r *MultilineSplitter) Init(config interface{}) error {
+	conf := config.(*MultilineSplitterConfig)
+	var err error
+	if r.delimiter, err = regexp.Compile(conf.Delimiter); err != nil {
+		return err
+	}
+	if r.delimiter.NumSubexp() > 1 {
+		return fmt.Errorf("regex must not contain more than one capture group: %s",
+			conf.Delimiter)
+	}
```
rafrombrc: Not sure why this constraint is here. In the RegexSplitter, the capture group specifies which part of the regex match, if any, should be included in the record that is emitted. AFAICT you're always throwing the delimiter away, so this has no impact.

Ultimately, it would probably be better to match the behaviour of the RegexSplitter, to enable people to keep the delimiter in the record if desired.

pipeline/splitters.go
relistan: Yeah, agreed. I called it out in the description of the PR, but it's easy enough to make it configurable.

pipeline/splitters.go
relistan: Hmm, good catch. I think that's a leftover from some code that was in the `RegexSplitter`, which has this restriction. The reason I don't allow you to configure whether or not you can keep the delimiter is that it will always keep all the intervening delimiters–it doesn't make sense otherwise. But I suppose it might still make sense to configure whether or not you keep the last one. I'll just remove the restriction.

relistan:
Excellent. I don't currently have the time allocated to update the other two splitters, but I'd be willing to do it on a future pass and will plan to allocate that time in the next project. As for naming... this name really comes from Logstash and the way it handles multiline stuff. So I think it probably makes sense to people from that arena. But it should not just make sense to some people, it needs to not confuse anyone! How about something like `PatternGroupingSplitter`?

I'll update the docs and fix the above commented sections and push, probably early next week.

rafrombrc:
I can live with either MultiPatternSplitter or PatternGroupingSplitter, w/ slight pref for the first, as long as the use case is clearly explained in the docs.

relistan:
Code is updated, will work on docs.

relistan:
@rafrombrc docs are added, ready for review.

relistan:
@rafrombrc I know you don't have much time for this these days (been there with my own open source projects). Would very much appreciate a review if you can get the time.

PR #1843, opened 2016-01-26
A major revamp of the DockerLogInput plugin

relistan:
This PR is a general cleanup of the Docker input plugin to accomplish three major goals:

1. Don't lose connections to Docker when the daemon is down for a period of time
2. Support the splitter API in this plugin
3. Log what is happening

From running Heka with Docker for quite some time at my previous employer (New Relic) we saw that Heka would lose connections to the Docker daemon if it was down for a period of time (e.g. not a restart but a full stop for minute). This was because the re-connect strategy was to restart the plugin entirely and Heka would give up doing that after a short time. This pR takes a different approach and tries very hard to stay connected to the Docker daemon while also still exiting the plugin on total failure.

Secondly a major drawback of the existing plugin is that lack of support for the new splitters means that you are stuck with with-by-line output from this Input. This PR supports splitters in the `DockerLogInput` plugin. At Nitro we're now using this with our `MultilineSplitter` implementation in #1842 to grab stacktraces that are output from apps running under Docker.

Finally, without logging, it's really hard to tell what the plugin is seeing an not seeing. This adds some minimalist logging that seems to be enough to tell when things are attached properly or not.

### DIFF
--- a/plugins/docker/attacher.go
+++ b/plugins/docker/attacher.go
@@ -24,12 +24,23 @@ package docker
 
 import (
 	"bufio"
+	"fmt"
 	"io"
 	"path/filepath"
 	"strings"
 	"sync"
+	"time"
 
-	"github.com/carlanton/go-dockerclient"
+	"github.com/fsouza/go-dockerclient"
+	"github.com/mozilla-services/heka/pipeline"
+)
+
+const (
+	CONNECT_RETRIES = 4
+)
+
+var (
+	retryInterval = []time.Duration{1, 2, 3, 7}
 )
 
 type AttachEvent struct {
@@ -44,31 +55,23 @@ type Log struct {
 	Fields map[string]string
 }
 
-type Source struct {
-	ID     string
-	Name   string
-	Filter string
-	Types  []string
-}
-
-func (s *Source) All() bool {
-	return s.ID == "" && s.Name == "" && s.Filter == ""
-}
-
 type AttachManager struct {
 	sync.RWMutex
 	attached      map[string]*LogPump
 	channels      map[chan *AttachEvent]struct{}
 	client        DockerClient
-	errors        chan<- error
 	events        chan *docker.APIEvents
 	eventReset    chan struct{}
 	sentinel      struct{}
+	ir            pipeline.InputRunner
+	endpoint      string
+	certPath      string
 	nameFromEnv   string
 	fieldsFromEnv []string
 }
 
-func NewAttachManager(endpoint, certPath string, attachErrors chan<- error, nameFromEnv string, fieldsFromEnv []string) (*AttachManager, error) {
+// Return a properly configured Docker client
+func newDockerClient(certPath string, endpoint string) (DockerClient, error) {
 	var client DockerClient
 	var err error
 
@@ -81,6 +84,14 @@ func NewAttachManager(endpoint, certPath string, attachErrors chan<- error, name
 		client, err = docker.NewTLSClient(endpoint, cert, key, ca)
 	}
 
+	return client, err
+}
+
+// Construct an AttachManager and set up the Docker Client
+func NewAttachManager(endpoint string, certPath string, nameFromEnv string,
+	fieldsFromEnv []string) (*AttachManager, error) {
+
+	client, err := newDockerClient(certPath, endpoint)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


relistan:
I should have pointed out that there is one other piece of new functionality here: this supports additional fields being pulled from container labels. These were added in Docker 1.6 and are the best practice way to store metadata about containers. The config directive is called `fields_from_labels` and it works basically identically to `fields_from_env` but using labels as the source.

plugins/docker/attacher.go
```diff
@@ -81,52 +70,142 @@ func NewAttachManager(endpoint, certPath string, attachErrors chan<- error, name
...
 
-	// Attach to all currently running containers
-	if containers, err := client.ListContainers(docker.ListContainersOptions{}); err == nil {
-		for _, listing := range containers {
-			m.attach(listing.ID[:12])
+	return m, nil
+}
+
+// Handler to wrap functions with retry logic
+func withRetries(doWork func() error) error {
```
rafrombrc: We've got a [RetryHelper](https://github.com/mozilla-services/heka/blob/dev/pipeline/retry_helper.go) that does exponential back-off with jitter that we use elsewhere throughout Heka. Someday we might even add support for the helper noticing when a shutdown is happening to abort without latency. Regardless, it'd be great if you could use that in here rather than hand-rolling.

plugins/docker/attacher.go
```diff
@@ -81,52 +70,142 @@ func NewAttachManager(endpoint, certPath string, attachErrors chan<- error, name
...
+	// Retry this up to CONNECT_RETRIES number of times, sleeping
+	// the defined interval. During this time the Docker client should
+	// get reconnected if there is a general connection issue.
+	err := withRetries(m.attachAll)
+	if err != nil {
+		m.ir.LogError(err)
+		m.ir.LogError(
+			fmt.Errorf("Failed to attach to Docker containers after %s retries. Plugin giving up.", CONNECT_RETRIES),
+		)
+		return err
```
rafrombrc: Here and in a couple of other places it looks like AttachManager.Run is logging an error and then returning the error, but then the input itself returns the error, which will cause Heka to log the error a second time.

plugins/docker/attacher.go
```diff
@@ -81,52 +70,142 @@ func NewAttachManager(endpoint, certPath string, attachErrors chan<- error, name
...
+	}
+
+	m.ir.LogMessage("Attached to all containers")
+
+	err = withRetries(func() error { return m.client.AddEventListener(m.events) })
+	if err != nil {
+		m.ir.LogError(
+			fmt.Errorf("Failed to add Docker event listener after %s retries. Plugin giving up.", CONNECT_RETRIES),
+		)
+		return err
```
rafrombrc: Oops, I misread... you're *not* returning `err` here, so there won't be duplicate logging. But the order will likely be a bit confusing. I'd recommend logging the `err` returned from `withRetries` and then returning the `giving up` message so it shows up afterwards. Same with above.

plugins/docker/attacher.go
```diff
@@ -81,52 +70,142 @@ func NewAttachManager(endpoint, certPath string, attachErrors chan<- error, name
...
+	return nil
+}
+
+// Try to reboot all of our connections
+func (m *AttachManager) restart() {
+	var err error
+
+	m.ir.LogMessage("Lost connection to Docker, re-connecting")
+	m.client.RemoveEventListener(m.events)
+	m.events = make(chan *docker.APIEvents) // RemoveEventListener closes it
```
rafrombrc: This comment is misleading, RemoveEventListener doesn't actually close the channel, it only removes it from the set of listeners.

plugins/docker/attacher.go
```diff
@@ -81,52 +70,142 @@ func NewAttachManager(endpoint, certPath string, attachErrors chan<- error, name
...
+
+	m.client.AddEventListener(m.events)
+
+	err = withRetries(m.attachAll)
+	if err != nil {
+		m.ir.LogError(fmt.Errorf("Unable to attach to some containers! (%s)", err.Error()))
```
rafrombrc: I'm a bit confused. If `withRetries(m.attachAll)` returns an error in the Run method, then Run itself returns the error which AFAICT will cause the plugin to exit. But if it returns an error here in the restart method, then we just log the error and take no further action, everything keeps running as usual. Am I understanding this correctly? Is this actually the behaviour that we want?

plugins/docker/attacher.go
```diff
@@ -137,12 +216,36 @@ func (m *AttachManager) attach(id string) {
...
+}
+
+// Attach to the output of a single running container
+func (m *AttachManager) attach(id string, client DockerClient) error {
+	m.ir.LogMessage(fmt.Sprintf("Attaching container: %s", id))
+
+	fields, err := m.extractFields(id, client)
+	if err != nil {
+		m.ir.LogError(err)
+		return err
```
rafrombrc: There are a lot of places where an error is logged before being returned. With the `withRetry` decorator pattern, it's a bit hard to keep track of where all of the error logging happens, but I'm pretty sure that there are more duplicates than just the ones already mentioned. Generally I recommend pushing the logging to the topmost layer that touches the error.

plugins/docker/attacher.go
```diff
@@ -152,165 +255,93 @@ func (m *AttachManager) attach(id string) {
 			Stream:       true,
 			Success:      success,
 		})
+
+		// Once it has exited, close our pipes
 		outwr.Close()
 		errwr.Close()
 		if err != nil {
+			m.ir.LogError(err)
```
rafrombrc: And again w/ logging ambiguity.

plugins/docker/attacher.go
relistan: Great, hadn't seen anything using that so used a pattern I've used a lot. I'll use the `RetryHelper` instead.

plugins/docker/attacher.go
relistan: Yeah, good call. It looks like when I refactored things to delegate `Run` to the `AttachManager` it now logs this twice. Will fix.

plugins/docker/attacher.go
relistan: Sure.

plugins/docker/attacher.go
relistan: Well, the reason that comment is there is that `RemoveEventListener` does (sometimes) also close it. Experience showed that it always closed the channel. But in double checking this I just discovered that it only closes it if there are no more event listeners, otherwise it appears to leave it open from what I can tell. I actually need to check that here so we don't leak open channels.

rafrombrc: Not only a leak, but in handleDockerEvents we call restart if m.events is closed, and then restart calls `RemoveEventListener(m.events)`. If there's a case where RemoveEventListener might again call close, then we'll panic.

rafrombrc:
First, thanks *so* much for your contribution, on this and the other PR, your work is very much appreciated. Overall this code looks great; I've made a bunch of comments inline, but as you can see they're mostly smaller issues, although a couple of them might have a bigger functional impact.

I have one more monkey-wrench to throw into the mix, however. I've just recently gotten a request from inside Mozilla to investigate the possibility of getting the DockerLogInput to be able to maintain its location in the containers' log streams, so Heka can shut down and restart without necessarily losing log records from the container. The Docker API's `logs` endpoint, supported by fsouza's Go client, provides the `since` parameter that could be used for this purpose.

I don't necessarily expect you to tackle that, but if it's something that would be useful to you as well and you feel like working on it while you're in there, then support for using `logs` instead of `attach` would be awesome. If not, then you can just clean up the work you've already done based on my comments, we'll merge it, and treat the stateful log tracking as a separate issue.

Oh, one more note: it'd be great if you could add a record to the CHANGES.txt changelog for the v0.11 release. Thanks!

plugins/docker/attacher.go
relistan: You are correct about how it works. This should be returning the error and then breaking out of the loop in `handleDockerEvents` though. That would cause the plugin to exit, which is what I think we want it to do here.

plugins/docker/attacher.go
relistan: That's reasonable. There was no error logging before, so hey, it's better than that. I'll take a pass through it and consider where it might be duplicated.

relistan:
Thanks, Rob. Yeah I had the same thought about the logging position. Right now you just miss whatever logs went by while Heka wasn't listening. Being able to capture them all would be ideal. Is there any existing state mechanism for Heka plugins (i.e. between Heka restarts)?  I think I'll not tackle that right now, but will think it over because I think it's important. I have another project going at the moment, but when it's complete I'll be back to this. Likely Monday, but maybe tomorrow if I'm quick. ;)

plugins/docker/attacher.go
relistan: Hmm, yeah that's rather nasty. I think this is actually a bug in the `go-dockerclient` because it should either do one or the other but not sometimes one and sometimes the other. I'll work around it here, but I'll open a bug there.

intjonathan:
:heart_eyes:

plugins/docker/attacher.go
rafrombrc: To be clear, I think the existence of the withRetries decorator is fine, as long as it uses RetryHelper internally.

relistan: Done.

relistan:
@rafrombrc Ok, this is ready for re-review. I took into account all your suggestions, and I went through all the logging to see where and when messages were being generated. I removed some calls and added a couple of others where it was falling short. I added the changes to `CHANGES.txt` as requested as well.

relistan:
@rafrombrc thanks!

PR #1844, opened 2016-01-27
Update HttpInput docs for version 0.10

gjtempleton:
Updates docs for HTTP Input plugin to correct input name.

Also moves misleadingly placed headers subsection to bottom of list. (Though this is largely personal preference.)

### DIFF
--- a/docs/source/config/inputs/http.rst
+++ b/docs/source/config/inputs/http.rst
@@ -50,14 +50,9 @@ Config:
 - headers (subsection):
     .. versionadded:: 0.5
 
-    Subsection defining headers for the request. By default the User-Agent
-    header is set to "Heka"
-- body (string):
-    .. versionadded:: 0.5
-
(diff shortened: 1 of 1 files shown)


docs/source/config/inputs/http.rst
```diff
@@ -80,6 +75,11 @@ Config:
     Severity level of errors, unreachable connections, and non-200 responses
     of successful HTTP requests. Defaults to 1 (alert).
 
+    Subsection defining headers for the request. By default the User-Agent
+    header is set to "Heka"
+- body (string):
+    .. versionadded:: 0.5
+
```
rafrombrc: Looks like you were trying to move the `headers`  piece down here to the bottom, but you instead moved the headers description and the following `body` label.

gjtempleton: Yep, exactly what I was trying to do. Realised at the time. Apparently my push to fix it failed, pushed now.

PR #1845, opened 2016-01-27
Update parameters for HttpInput plugin to match HttpListen and docs

gjtempleton:
This updates the HttpInput plugin's config parameters to match the HttpListenInput plugin and the docs. 

See this for updating v0.10.0 docs to match current naming: [https://github.com/mozilla-services/heka/pull/1844]

### DIFF
--- a/cmake/externals.cmake
+++ b/cmake/externals.cmake
@@ -175,7 +175,7 @@ if (INCLUDE_GEOIP)
 endif()
 
 if (INCLUDE_DOCKER_PLUGINS)
-    git_clone(https://github.com/carlanton/go-dockerclient d408f209d5946d86da69382b3eb0a6faac7b3885)
+    git_clone(https://github.com/fsouza/go-dockerclient 175e1df973274f04e9b459a62cffc49808f1a649)
 endif()
 
 if (INCLUDE_MOZSVC)
--- a/cmake/externals.cmake
+++ b/cmake/externals.cmake
@@ -175,7 +175,7 @@ if (INCLUDE_GEOIP)
 endif()
 
 if (INCLUDE_DOCKER_PLUGINS)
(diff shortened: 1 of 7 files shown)


cmake/externals.cmake
```diff
@@ -175,7 +175,7 @@ if (INCLUDE_GEOIP)
 endif()
 
 if (INCLUDE_DOCKER_PLUGINS)
-    git_clone(https://github.com/carlanton/go-dockerclient d408f209d5946d86da69382b3eb0a6faac7b3885)
+    git_clone(https://github.com/fsouza/go-dockerclient 175e1df973274f04e9b459a62cffc49808f1a649)
```
rafrombrc: This is unrelated and shouldn't be included in this PR.

plugins/docker/attacher.go
```diff
@@ -29,7 +29,7 @@ import (
 	"strings"
```
rafrombrc: Everything in this file is also unrelated and shouldn't be included in this PR.

plugins/docker/client.go
```diff
@@ -1,6 +1,6 @@
 package docker
 
-import "github.com/carlanton/go-dockerclient"
+import "github.com/fsouza/go-dockerclient"
```
rafrombrc: Same here and in the next file, none of the docker stuff is relevant to this PR.

rafrombrc:
In addition to removing the unrelated docker changes that got included in this PR, it needs an entry in the "Backwards Incompatibilities" section of the CHANGES.txt changelog for the 0.11.0 release.

Thanks!

gjtempleton:
The perils of making changes and a PR at the end of a long, long work day...

Sorry, really should have picked up on that. Fixed now.

PR #1848, opened 2016-02-01
add payload_keep option to nginx_error.lua decoder

relud:


relud:
I need to be able to preserve payload on all logs, so that I can extract it later for creating aggregated versions of the original logs.

PR #1852, opened 2016-02-04
Add a csv decoder

nacerix:
Add a csv decoder working in a similar way as payload xml decoder.
Fields names are mapped to their position in records and then interpolated in message variables.


rafrombrc:
While I very much appreciate your contribution, and especially that you were thoughtful enough to include tests, I'm afraid I'll have to pass on this PR. Nearly all text parsing in Heka is done using Lua and LPEG in SandboxDecoders. Our LPEG test site even has a simple [example](http://lpeg.trink.com/share/csv) of CSV parsing.

Issue #1841, opened 2016-01-26
Add a configuration option to the logstreamer input for the "check for more data" interval


simonpasquier:
As noted by @elemoine on the [mailing list](https://mail.mozilla.org/pipermail/heka/2016-January/000947.html), a single logstreamer instance monitoring many files may increase dramatically the CPU load of the system. This is because the logstreamer plugin checks every 250ms if there's more data to collect and the value is hardcoded.

The idea would be have a configuration option to set the "check for more data" interval with a default value of 250ms.

aleksmm:
in my config in #1626 I have single log file in logstreamer config to listen to and still have same issue. If this matters.

PR #1862, opened 2016-02-23
Add check_data_interval option for the LogStreamer input

simonpasquier:
This change addresses the issue #1841. To avoid excessive CPU usage,
users can set a value higher than the default 250ms when they know that
a single logstreamer input may process many log files.


simonpasquier:
I forgot to update the documentation. I'll send an updated PR soon.

PR #1867, opened 2016-02-29
Add a remap_original_input_fields configuration option for JSON decoder.

mattmoyer:
This allows fields from the original message to be renamed and attached to the decoded messages. This is useful to capture fields set in input plugin pack decorators, e.g, capturing `RemoteAddr` from the HttpListenInput or capturing `Offset` from the KafkaInput.

The syntax is a little funky (space-delimited list of `newField=newField`). I didn't see an easy way to pass a map/table configuration value from the configuration `.toml` down into a `SandboxPlugin`. Is this a reasonable way to do this?

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -38,6 +38,8 @@ Features
 
 * Added `fields_from_labels` config option to DockerLogInput (#1843).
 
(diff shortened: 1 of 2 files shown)


fire:
As someone who's not on the development team but would like to try this, can you update this to master?

PR #1870, opened 2016-03-01
plugin/http: Only discard response body if no error occured

nfnty:


rafrombrc:
I've cherry picked this commit (and added a changelog entry) onto the versions/0.10 branch so the bug fix exists there as well. Thanks for your contribution!

rafrombrc:
(see 3473f50a6cdd39d7d69eec16e870ea8fdd652df1)

Issue #1809, opened 2015-12-07
Add `separator` option to JsonDecoder


sbward:
JsonDecoder needs an option to change the separator character that's used to join nested keys together.  Currently the separator is hard-coded to be a dot (`.`).

This is important for users of Elasticsearch 2.0 because dots are no longer allowed in key names.  For that use case it's necessary to change the separator to something else, like an underscore.

At the time of this posting, JsonDecoder is implemented here: https://github.com/mozilla-services/heka/blob/dev/sandbox/lua/decoders/json.lua -- and the function that joins the keys together is here: https://github.com/mozilla-services/lua_sandbox/blob/35e2c211c0dcbe37c8a686033b4a57e18a28711c/modules/util.lua#L17

I recommend adding an optional argument to override the character and exposing it in JsonDecoder's option set.

Thank you! 😎

ei-grad:
Imho it would be better if Elasticsearch JSON Encoder would translate nested keys into nested JSON objects.

roddik:
Any updates on this? This issue renders heka unusable for any new setup, that logs with nested json, and uses up-to-date elasticsearch. Logstash has a dedot filter for this, and adding a new option as suggested by sbward is easy.

nfnty:
I implemented `ESJsonEncoder` `field_name_substitute` for this: https://github.com/nfnty/heka/commit/7540517254371c8d17eda2020226481affcf6aec

I may send a PR later.

rafrombrc:
Relevant: mozilla-services/lua_sandbox#118.

PR #1872, opened 2016-03-03
Update lua_sandbox, add max-depth and sep-char to JsonDecoder

nathwill:


rafrombrc:
Closes #1809.

nathwill:
@rafrombrc i'm not sure why exactly, but there does seem to be something weird with the nesting... the test matches experimental results but `GetFieldValue("nested1-field")` keeps getting `nil`, so those 2 tests are failing :crying_cat_face:, still digging, but if you have any pointers, i'd appreciate it.

rafrombrc:
Test was failing b/c sandbox config expects number values to be type int64 or float64, unspecified is type int. I fixed this, changed the name of the setting from `separator_char` to `separator`, and added a changelog entry. Merging now. Thanks for tackling this!

nathwill:
awesome, thanks Rob!

Issue #1452, opened 2015-03-31
LogstreamerInput mishandling file rotation


rafrombrc:
When a file rotation scheme has the current file being rolled over into a new file that didn't exist before the rotation, LogstreamerInput is not correctly tracking the rotation. Detailed information on the initial report of this issue can be found in this mailing list thread: https://mail.mozilla.org/pipermail/heka/2015-March/000409.html

rajivek:
Seems current fix in logstreamer_rotate_fix` branch is working. But when we doing a long run its not properly reading the logs (i.e missing some data too). 

I have used only one single log stream for testing but when we doing its running for around 20 - 30 min heka give "too many files open error". Currently its rotating in every 5 min.  

http://pastebin.ca/2963610 - toml file

http://pastebin.ca/2963598 - sh script for simulate the log rotation.

same script increasing the number of lines per log

rafrombrc:
Hrm... in my testing I'm seeing some file descriptor usage, but nowhere near enough to come close to the fd limit. How many files exist in the file stream?

rafrombrc:
I'd still like to know how many files exist in your file stream, but I've also added another commit to the logstreamer_rotate_fix branch that might help if my guess that there are a *lot* (i.e. thousands) of files is correct. Can you pull down the latest changes, recompile, and report back to let me know if it helps? Thanks.

rajivek:
I have tested this issue with many files, I did not find any issues after your commit. Any way need to check with long run.

rajivek:
I have tested this.. fix.. I haven't see any issue with this fix. Please merge this with the release.

Thank you for the help

guluglu:
This question is still exist in version 0.11.0

PR #1874, opened 2016-03-04
Merge changes from 0.9 into 0.10

bjornvester:
The fix for the log rotation issue #1452 was merged into the 0.9 branch and was supposed to be part of a 0.9.3 release. From what I can see, a 0.10 branch was created instead and these fixes were not merged over.

We have been hit with the log rotation issue in the 0.10 release and can reproduce the problem. I have tested these changes and I can confirm that they fix the problem at least for us.


rafrombrc:
Thanks so much!

PR #1876, opened 2016-03-06
make http_status filter status field configurable

nathwill:


rafrombrc:
I'll go ahead and merge this, but I'll note that, as much as possible, decoders should be written so that when a message represents an HTTP request it uses the default schema expected by the http_status filter. It's meant to be a de facto standard for how HTTP requests should be represented in Heka message format.

nathwill:
thanks rob! hua.

PR #1877, opened 2016-03-08
Fix DockerEvents input

gjtempleton:
Fixes argument ordering of call to newDockerClient

@relistan FYI: Seems your update changed the order of the args in `newDockerClient` from `(Endpoint, CertPath)` to `(CertPath, Endpoint)`


relistan:
Good catch, thanks!

PR #1880, opened 2016-03-09
Fix typo in the configuration documentation

simonpasquier:


elemoine:
Good catch. Looks good to me :)

PR #1888, opened 2016-03-16
Add iowait percentage in filters/procstat.lua

McStork:
Hi,
It would be cool to have iowait as a percentage. Pretty important stat.


sathieu:
:+1:

PR #1896, opened 2016-03-29
Fix incorrect timestamp

actaeon:
Heka message expects nanoseconds since the UNIX epoch


rafrombrc:
Oops, good catch. Got bit by an overzealous search-and-replace. Thanks.

PR #1917, opened 2016-04-27
Update installing.rst: replaced 'source' with '.'

ekalinin:
`source` is synonymous for `.` in bash and zsh only.
Some lighter shells don't have `source` command.
Moreover, only `.` is specified by POSIX.
So ksh, dash, ash, etc. may not have `source`.

### DIFF
--- a/docs/source/installing.rst
+++ b/docs/source/installing.rst
@@ -70,7 +70,7 @@ Build Instructions
     .. code-block:: bash
 
         cd heka
-        . build.sh # Unix (or `. build.sh`; must be sourced to properly setup the environment)
+        . build.sh # Unix
         build.bat  # Windows
 
 You will now have a ``hekad`` binary in the ``build/heka/bin`` directory.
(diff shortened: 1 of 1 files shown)


docs/source/installing.rst
```diff
@@ -70,7 +70,7 @@ Build Instructions
     .. code-block:: bash
 
         cd heka
-        source build.sh # Unix (or `. build.sh`; must be sourced to properly setup the environment)
+        . build.sh # Unix (or `. build.sh`; must be sourced to properly setup the environment)
```
iandioch: The comment here doesn't make much sense anymore

rafrombrc:
Thanks for your contribution, but I'm going to leave it as is. We used to just have `. build.sh`, as you do, but we ended up w/ people commonly misreading it, using `./build.sh`, and showing up in IRC asking why it wasn't working. We changed it to `source build.sh` while still including a reference to `. build.sh` for the folks who aren't using bash and we haven't had any confusion since.

PR #1921, opened 2016-04-29
Add ability to deflate gzip'd payloads in HttpListenInput

hynd:
Useful for uncompressing data from something like http://bosun.org/scollector/ or other gzip'd payloads that need splitting before decoding?

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -51,6 +51,8 @@ Features
 
 * Added iowait percentage output field in filter procstat (#1888).
 
+* Added ability to deflate gzip'd payloads in HttpListenInput.
+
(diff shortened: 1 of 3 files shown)


bigkraig:
:+1: although why make this configurable? is the data useful if its still compressed?

bigkraig:
@rafrombrc @trink Whats the likelihood of this being included? I am evaluating using heka to funnel scollector metrics, which are posted in gzip.

hynd:
The optional/opt-in-ness was in case some already use Heka just as a proxy for compressed HTTP payloads to another upstream (happy to change the behaviour!)

PR #1924, opened 2016-05-04
Graylog native go plugin

xrl:
I needed to get logs out of the docker daemon and the most attractive option was graylog. Problem is that my version of docker (1.9) could not disable compression and the simple graylog decoder was not going to be sufficient. So I tackled the native plugin. This should be a useful addition for other heka/docker users.

### DIFF
--- a/plugins/graylog/graylog_input.go
+++ b/plugins/graylog/graylog_input.go
@@ -9,7 +9,6 @@ import (
 
 type GraylogInputConfig struct{
 	Address string `toml:"address"`
-	Type string `toml:"type"`
 }
 
 type GraylogInput struct {
@@ -58,9 +57,9 @@ func (g *GraylogInput) Run(ir pipeline.InputRunner, h pipeline.PluginHelper) (er
 					break
 				}
 			}
-
-			close(g.ctrlMsgs)
 		}
+		close(g.ctrlMsgs)
+
 	}()
 
 	for ctrlMsg := range g.ctrlMsgs {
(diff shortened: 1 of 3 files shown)


plugins/graylog/graylog_input.go
```diff
@@ -0,0 +1,120 @@
...
+	err error
+	message *gelf.Message
+}
+
+func (g *GraylogInput) Run(ir pipeline.InputRunner, h pipeline.PluginHelper) (err error) {
+	go func() {
+		for {
+			select {
+			case <-g.stopChan:
+				break
```
rafrombrc, line 53: This break statement (and the one below on L61) will only break from the select, which means that AFAICT the for loop will never exit.

plugins/graylog/graylog_input.go
```diff
@@ -0,0 +1,120 @@
...
+			default:
+				message,err := g.reader.ReadMessage()
```
rafrombrc, line 55: I'm not positive, but from looking at the graylog client code it looks like this call might block forever, which means Heka shutdown would hang if nothing was coming over the connection.

plugins/graylog/graylog_input.go
```diff
@@ -0,0 +1,120 @@
...
+				g.ctrlMsgs <- gelfCtrl {
+					err: err,
+					message: message,
+				}
+				if err != nil {
+					break
+				}
+			}
+		}
+		close(g.ctrlMsgs)
```
rafrombrc, line 65: AFAICT this line will never be hit, but maybe I'm missing something?

plugins/graylog/graylog_input.go
```diff
@@ -0,0 +1,120 @@
...
+				ir.LogError(err)
+				break MsgLoop
+			}
+			pack.Message.AddField(field)
+		}
+
+		ir.Deliver(pack)
+	}
+
+	return
```
rafrombrc, line 105: I know there are plenty of examples of this littered throughout Heka code, but it's actually not recommended to use bare returns except for very short functions where the return var declaration is clearly visible from the return call, if then.

rafrombrc:
I know I said that I was open to including this in the Heka core, but I'm having second thoughts. I don't really want to add additional dependencies unless it's for an extremely common use case. I'd rather see this packaged up as a separate package that can easily be included in a Heka build using the plugin_loader.cmake stuff.

Also, for future reference, if a plugin is to be included in the Heka core, it needs to go in the changelog, and the documentation needs to be updated.

xrl:
Thanks for the feedback! I think graylog is a popular format, it's included as one of the default log drivers in the docker daemon (which is why I wrote this) and it's a default plugin in logstash.

There's already a graylog decoder in heka core but it's not broadly useful because it can't do the standard decompression you'll see coming out of graylog connections.

If it's any consolation, the included dependency is a very simple one and does not itself include other dependencies. It makes heavy use of the go stdlib.

PR #1926, opened 2016-05-04
Beefed up .travis.yml so we can get automatic builds.

xrl:
Generate docker images after successful build and push up to the central docker registry. We'll need the heka project to create an account on the registry and then inject credentials for this travis to work.

You'll need to install the travis command line tool and inject some docker variables:

```
gem install travis
travis login --org
travis env set DOCKER_EMAIL $YOUR_DOCKER_ACCT_EMAIL
travis env set DOCKER_USERNAME $YOUR_DOCKER_ACCT_USERNAME
travis env set DOCKER_PASSWORD $YOUR_DOCKER_ACCT_PASSWD
travis env set --public DOCKER_REPO_SLUG mozilla-services/heka
```

These values are used by the `docker/release_travis.sh`. By default `env set` creates private variables which don't show up in the build logs. The `DOCKER_REPO_SLUG` is made public so it can be clear where the image is going. The unit tests must pass for the docker image to be built.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,13 +6,16 @@ notifications:
     irc:
         channels:
             - "irc.mozilla.org#heka"
+services:
+    - docker
 addons:
     apt:
         packages:
         - protobuf-compiler
         - cmake
         - libgeoip-dev
+        - debhelper
 install:
     - . build.sh
 script:
-    - make test
+    - make test && make deb && ./../docker/release_travis.sh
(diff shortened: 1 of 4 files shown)


xrl:
This is failing because the travis account has not had `travis env set --public DOCKER_REPO_SLUG mozilla-services/heka`. Once this `DOCKER_REPO_SLUG` is set it will be available to all travis builds -- and it would just make sense to kick Travis to build again.

PR #1930, opened 2016-05-11
Add heartbeat filter doc

sathieu:


sathieu:
Same as #1923, for v0.10

PR #1944, opened 2016-06-07
Catch when stats channel is closed

gjtempleton:
- Discovered during use that removal of a container the plugin has been attached to causes a panic due to attempting to close the statsrd channel when already closed by the go-dockerclient. This fix means it no longer attempts to do this and ensures it returns from the gofunc when it is closed.

I should have caught this sooner and only found it when using it in anger. Pretty sure this is the best way to solve the issue but any feedback is appreciated as usual.


relistan:
👍

PR #1916, opened 2016-04-26
BIND query log decoder Lua plugin

nickchappell:
@rafrombrc @trink do you want this squashed into a single commit?

PR #1945, opened 2016-06-08
Move the grammar portion of the BIND query log decoder script into it…

nickchappell:
…s own module, and make the decoder script reference it.

@trink @rafrombrc Merging #1916 after I modularized the decoder caused it to break and crash Heka on startup.

This PR keeps the modularized grammar and moves it to `modules/` so it's usable elsewhere and re-adds the actual usable decoder script.

Would you like a PR against https://github.com/mozilla-services/lua_sandbox/ to add the `bind.lua` module there as well?


sandbox/lua/decoders/bind_query_log.lua
```diff
@@ -48,197 +45,45 @@ local date_time = require 'date_time'
...
---and followed by one more hostname_fragment, indicating the end of a complete
---hostname. The open and close parens and colon are to match the decorations
---BIND puts around the name: 
--- (webserver.company.com):
-local enclosed_query = "(" * l.Cg((hostname_fragment * ".")^1 * hostname_fragment, "FullQuery") * "):"
+  --Create a local variable for the log line from the Payload of the incoming Heka message.
+  --The LogstreamerInput that generates the messages automatically puts the 
+  local query_log_line = read_message("Payload")
+  --Use the bind_query grammar defined farther above to create a local grammar object:
+  local grammar = l.Ct(bind.query_log_grammar)
```
rafrombrc: This never changes, it can be set outside of the `process_message` function so it's not reinitialized for every message.

sandbox/lua/decoders/bind_query_log.lua
```diff
@@ -48,197 +45,45 @@ local date_time = require 'date_time'
...
+  fields = grammar:match(query_log_line)
 
---The ^-1 means match at most 1 instance of the pattern. We want this so that we 
---can match the first part of a hostname and leave the rest for the l.Cg((hostname_fragment...
---capture group to match into the QueryDomain table entry.
---In webserver.company.com, `(hostname_fragment * ".")^-1` matches webserver.
---and `l.Cg((hostname_fragment...` matches company.com
-local query = l.Cg((hostname_fragment)^-1, "QueryName") * "." * l.Cg((hostname_fragment * ".")^1 * hostname_fragment, "QueryDomain")
+  --If fields is empty, exit immediately:
+  if not fields then return -1 end
```
rafrombrc: You can return a second value containing an error message, I'd recommend it here.

sandbox/lua/decoders/bind_query_log.lua
```diff
@@ -48,197 +45,45 @@ local date_time = require 'date_time'
...
+  --Include the original, unparsed log line as the Payload of the message:
+  msg.Payload = read_message("Payload")
+  inject_message(msg)
 
--- To use the above grammar on http://lpeg.trink.com/, paste in everything above this line and just the single line below (uncomment it too):
---grammar = l.Ct(bind_query)
+  return 0
 
-return M
+end
```
rafrombrc: Add newline at the EOF.

sandbox/lua/modules/bind.lua
```diff
@@ -0,0 +1,226 @@
...
+--and `l.Cg((hostname_fragment...` matches company.com
+local query = l.Cg((hostname_fragment)^-1, "QueryName") * "." * l.Cg((hostname_fragment * ".")^1 * hostname_fragment, "QueryDomain")
+
+--Use all of the previously defined patterns to build a grammar:
+query_log_grammar = timestamp * space * queries_literal * space * info_literal * space * client_literal * space * client_ip * space * enclosed_query * space * query_literal * space * query * space * dns_record_class * space * dns_record_type * space * query_flags * space * server_responding_ip
+
+-- To use the above grammar on http://lpeg.trink.com/, paste in everything above this line and just the single line below (uncomment it too):
+--grammar = l.Ct(bind_query)
+
+return M
```
rafrombrc: Add newline at EOF.

rafrombrc:
Overall this looks pretty good. I've added a few comments. The only additional note I have is related to the docs. Currently you've got the docs in the comments of the Lua file, and then another copy of them in the docs folder. For the other sandbox plugins, we just pull the Lua file comments into the generated docs so we don't have to maintain two copies. You can look at the [apache access decoder page](https://raw.githubusercontent.com/nickchappell/heka/nickchappell/fix_bind_query_log_decoder_script/docs/source/config/decoders/apache_access.rst) to see how it works.

sandbox/lua/decoders/bind_query_log.lua
nickchappell: I didn't find any examples to follow in the Heka codebase but found one in `src/test/lua/lpeg_clf.lua` of the lua_sandbox: https://github.com/mozilla-services/lua_sandbox/blob/master/src/test/lua/lpeg_clf.lua

```
if not fields then
  error("no result", 2)
end
```

What is the intended meaning of the `expected` return value? The number of fields that were expected?

nickchappell:
@rafrombrc I made the recommended changes. On the doc issue, I removed the redundant example usage from the block comment in the Lua script, but will Sphinx pull in all of the block comments from the Lua file or just the first block comment?

sandbox/lua/decoders/bind_query_log.lua
rafrombrc: No, I'm not talking about the `error()` function, I'm talking about the `return -1`. You can return multiple values, and if you return a string for the second value it will be used as the error message. There are a couple of examples in the [json](https://github.com/mozilla-services/heka/blob/be1420d226808485a557ad622bb75785cadfa43a/sandbox/lua/decoders/json.lua#L175) [decoder](https://github.com/mozilla-services/heka/blob/be1420d226808485a557ad622bb75785cadfa43a/sandbox/lua/decoders/json.lua#L201).

rafrombrc:
The `bind_query_log.rst` file shouldn't contain any of the actual documentation, it should all be contained in the Lua comment. Sphinx pulls in all of the text in between two specially constructed (and explicitly specified) comment markers in the lua file. Please look carefully at the apache access log [doc source](https://raw.githubusercontent.com/mozilla-services/heka/dev/docs/source/config/decoders/apache_access.rst) and [lua code](https://github.com/mozilla-services/heka/blob/dev/sandbox/lua/decoders/apache_access.lua) to see what I mean.

nickchappell:
@rafrombrc I've made the requested changes to the error message and the rst/Lua docs. Do you want me to squash all of the commits before this can be merged?

rafrombrc:
Nope, this is fine. Thanks for your patience!

nickchappell:
@rafrombrc would you like a PR against the https://github.com/mozilla-services/lua_sandbox repo to add this there as well?

rafrombrc:
Sure. :+1:

Issue #1856, opened 2016-02-09
How to decode GET requests without body using HttpListenInput?


hypersleep:
Hi everyone! I tried to find something similar in issues list but i could not.

I have the following heka configuration:

```
[HttpListenInput]
address = "127.0.0.1:8080"

[RstEncoder]

[LogOutput]
message_matcher = "TRUE"
encoder = "RstEncoder"
```

When i try to send some GET requests without body - heka STDOUT is silent:

Request:
`$ curl -X GET localhost:8080/?qwe=asd`

But when i try to add some `zxc` payload to GET request:

`$ curl -X GET localhost:8080/?qwe=asd -d zxc`

I can see encoded messages in STDOUT with `zxc` payload field: 

```
2016/02/09 20:13:54
:Timestamp: 2016-02-09 17:13:54.194508833 +0000 UTC
:Type: heka.httpdata.request
:Hostname: airgun.local
:Pid: 17871
:Uuid: f426c66f-97c3-48ff-8221-7583a70944a1
:Logger: HttpListenInput
:Payload: zxc
:EnvVersion: 1
:Severity: 6
:Fields:
    | name:"Protocol" type:string value:"HTTP/1.1"
    | name:"UserAgent" type:string value:"curl/7.27.0"
    | name:"ContentType" type:string value:"application/x-www-form-urlencoded"
    | name:"Path" type:string value:"/"
    | name:"Host" type:string value:"localhost"
    | name:"RemoteAddr" type:string value:"127.0.0.1"
    | name:"qwe" type:string value:"asd"
```

Is there some way to decode messages without bodies?

Madhu1512:
I am also trying to parse url query string without body but no luck with HTTP GET. Looking for help on this 

+1

PR #1959, opened 2016-07-13
Decode GET requests without body

Madhu1512:
This PR is to parse querystring when payload is zero and also assign raw querystring to Query Field.

https://github.com/mozilla-services/heka/issues/1856


Madhu1512:
I will resubmit the PR after fixing the build issues.

PR #1972, opened 2016-08-02
Reorder some of the patterns in the dns_record_type capture group so …

nickchappell:
…that decode failures don't happen on records like A or NS where their names are identical to the first parts of other record names, like AAAA or NSEC.


nickchappell:
This is superseded by #1978.

